PR #4, opened 2022-05-27
Add autopep8 as a dev dependency

Lunarmagpie:
its nice to have


CircuitSacul:
perhaps also run this in the noxfile.py?

PR #6, opened 2022-05-27
Added config.json example

isFakeAccount:
Added the example for `config.json`. Hopefully, it is correct XD

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -1,3 +1,3 @@
 {
-  "discord_token": "asbdhjgasfgadsjhfgasdjfgkhsdjafgfa"
+    "discord_token": "asbdhjgasfgadsjhfgasdjfgkhsdjafgfa"
 }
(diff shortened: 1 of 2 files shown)


config.json.example
```diff
@@ -0,0 +1,3 @@
+{
+  "discord_token": "asbdhjgasfgadsjhfgasdjfgkhsdjafgfa"
```
CircuitSacul, line 2: will you indent with 4 spaces?

CircuitSacul (changes requested):

isFakeAccount:
pog

config.json.example
```diff
@@ -0,0 +1,3 @@
+{
+    "discord_token": "asbdhjgasfgadsjhfgasdjfgkhsdjafgfa"
```
Endercheif, line 2: instead of a random string of letters, perhaps make it either something more meaning full like `DISCORD-TOKEN` or a dummy token in the same format of a real token.

isFakeAccount: ok

CircuitSacul (approved):

PR #7, opened 2022-05-27
Make sure all non-empty files end in a newline

MithicSpirit:
Which one of you committed these sins (pun totally intended).


CircuitSacul:
My fault, I apologize

PR #18, opened 2022-07-07
Implement stats category

CircuitSacul:
Tested on testing server, works

### DIFF
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -19,12 +19,14 @@ class Bot(crescent.Bot):
 
         self._session: aiohttp.ClientSession | None = None
 
-    async def get_session(self) -> aiohttp.ClientSession:
-        if not self._session or self._session.closed:
-            self._session = aiohttp.ClientSession()
+    @property
+    def session(self) -> aiohttp.ClientSession:
+        if not self._session:
+            raise AttributeError("Session has not been set yet.")
         return self._session
 
     async def start(self, *args: Any, **kwargs: Any) -> None:
+        self._session = aiohttp.ClientSession()
         await super().start(*args, **kwargs)
 
         asyncio.create_task(loop_update_channels(self))
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -1,6 +1,13 @@
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+import aiohttp
 import crescent
 
 from mcodingbot.config import CONFIG
+from mcodingbot.tasks.stats import loop_update_channels
 
 
 class Bot(crescent.Bot):
@@ -9,3 +16,20 @@ class Bot(crescent.Bot):
 
         self.plugins.load("mcodingbot.plugins.info")
         self.plugins.load("mcodingbot.plugins.reactions")
+
+        self._session: aiohttp.ClientSession | None = None
+
+    async def get_session(self) -> aiohttp.ClientSession:
+        if not self._session or self._session.closed:
+            self._session = aiohttp.ClientSession()
+        return self._session
+
+    async def start(self, *args: Any, **kwargs: Any) -> None:
+        await super().start(*args, **kwargs)
+
+        asyncio.create_task(loop_update_channels(self))
+
+    async def join(self, *args: Any, **kwargs: Any) -> None:
+        await super().join(*args, **kwargs)
+        if self._session and not self._session.closed:
+            await self._session.close()
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -17,8 +17,8 @@ class Config:
     view_count_channel: int = -1
     member_count_channel: int = -1
 
-    mcoding_yt_id: str = ""
-    yt_api_key: str = ""
+    mcoding_yt_id: str = "YOUTUBE_CHANNEL_ID"
+    yt_api_key: str = "YOUTUBE_API_KEY"
 
     mcoding_youtube: str = (
         "https://www.youtube.com/channel/UCaiL2GDNpLYH6Wokkk1VNcg"
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -12,6 +12,14 @@ class Config:
     discord_token: str = "DISCORD_TOKEN"
     theme: int = 0x0B7CD3
 
+    mcoding_server: int = -1
+    sub_count_channel: int = -1
+    view_count_channel: int = -1
+    member_count_channel: int = -1
+
+    mcoding_yt_id: str = ""
+    yt_api_key: str = ""
+
     mcoding_youtube: str = (
         "https://www.youtube.com/channel/UCaiL2GDNpLYH6Wokkk1VNcg"
     )
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import logging
 import traceback
 from dataclasses import dataclass
 from math import log
@@ -10,6 +11,9 @@ from mcodingbot.config import CONFIG
 
 if TYPE_CHECKING:
     from mcodingbot.bot import Bot
+
+
+LOGGER = logging.getLogger(__file__)
 
 
 async def loop_update_channels(bot: Bot) -> None:
@@ -67,24 +71,31 @@ async def get_stats(bot: Bot) -> Stats:
         response: dict[str, Any] = await res.json()
 
     if not response:
+        LOGGER.error("Received 2XX but no data.")
         return _last_known_stats
 
     items = response.get("items")
     if not items:
+        LOGGER.error("Response did not contain 'items'.")
         return _last_known_stats
 
     channel = items[0]
     if channel.get("id") != CONFIG.mcoding_yt_id:
+        LOGGER.error("Channel ID was not mCoding.")
         return _last_known_stats
 
     statistics = channel.get("statistics")
     if not statistics:
+        LOGGER.error("mCoding channel did not contain 'statistics'.")
         return _last_known_stats
 
     subs = float(statistics.get("subscriberCount", 0))
     views = float(statistics.get("viewCount", 0))
 
     if not (subs and views):
+        LOGGER.error(
+            "Statistics did not contain subscriberCount and viewCount."
+        )
         return _last_known_stats
 
     _last_known_stats.subs = subs
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import traceback
 from dataclasses import dataclass
 from math import log
 from typing import TYPE_CHECKING, Any
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -36,13 +36,25 @@ async def update_channels(bot: Bot) -> None:
     assert view_channel
     assert member_channel
 
-    await sub_channel.edit(name=f"Subs: {display_stats(stats.subs)}")
-    await view_channel.edit(name=f"Views: {display_stats(stats.views)}")
+    if sub_channel:
+        await sub_channel.edit(name=f"Subs: {display_stats(stats.subs)}")
+    else:
+        LOGGER.warning("No sub count channel to update stats for.")
+
+    if view_channel:
+        await view_channel.edit(name=f"Views: {display_stats(stats.views)}")
+    else:
+        LOGGER.warning("No view count channel to update stats for.")
 
     guild = bot.cache.get_guild(CONFIG.mcoding_server)
-    assert guild
+    if not guild:
+        return LOGGER.warning(
+            "Couldn't find mCoding guild, not updating member count."
+        )
+
     member_count = guild.member_count
-    assert member_count is not None
+    if member_count is None:
+        return LOGGER.error("Cached guild didn't contain member count.")
 
     await member_channel.edit(name=f"Members: {display_stats(member_count)}")
 
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -47,7 +47,8 @@ async def update_channels(bot: Bot) -> None:
         return LOGGER.warning(
             "Couldn't find mCoding guild, not updating member count."
         )
-
+    if not member_channel:
+        return LOGGER.warning("No member count channel to update stats for.")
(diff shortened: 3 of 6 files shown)


mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@dataclass
+class Stats:
+    subs: float
+    views: float
+
+
+BASE_URL = "https://www.googleapis.com/youtube/v3/channels"
+LAST_STATS = Stats(0, 0)
```
Lunarmagpie, line 55: I feel like `MOST_RECENT_STATS` is a little clearer. Also i dont think this is a constant so it should be regular snake case.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+    pretty_stat = round(pretty_stat, 2)
+
+    exp_stat = round(log(int_stat, 2), 1)
+    # ^ this might not be as accurate as the member count thing when
+    # someone picky actually calculates it, but I suppose it's not
+    # gonna be such a problem if it's gonna be shown as e.g. "44.3K"
+
+    if exp_stat % 1 == 0:
+        exp_stat = int(exp_stat)
```
Lunarmagpie, line 115: shouldn't round already return an int?

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if pretty_stat % 1 == 0:
+        pretty_stat = int(pretty_stat)
```
Lunarmagpie, line 118: same here

mcodingbot/config.py
```diff
@@ -12,6 +12,14 @@ class Config:
     discord_token: str = "DISCORD_TOKEN"
     theme: int = 0x0B7CD3
 
+    mcoding_server: int = -1
+    sub_count_channel: int = -1
+    view_count_channel: int = -1
+    member_count_channel: int = -1
+
+    mcoding_yt_id: str = ""
+    yt_api_key: str = ""
```
Lunarmagpie, line 21: Maybe put something like `YT_API_KEY` (like `DISCORD_TOKEN` a few lines up)

mcodingbot/config.py
```diff
@@ -12,6 +12,14 @@ class Config:
     discord_token: str = "DISCORD_TOKEN"
     theme: int = 0x0B7CD3
 
+    mcoding_server: int = -1
+    sub_count_channel: int = -1
+    view_count_channel: int = -1
+    member_count_channel: int = -1
```
Lunarmagpie, line 18: I may be blind, but where are these being updated?

mcodingbot/bot.py
```diff
@@ -9,3 +16,20 @@ def __init__(self) -> None:
 
         self.plugins.load("mcodingbot.plugins.info")
         self.plugins.load("mcodingbot.plugins.reactions")
+
+        self._session: aiohttp.ClientSession | None = None
```
Lunarmagpie, line 20: I dont like this in bot. YouTubeAPI should be its own object.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+BASE_URL = "https://www.googleapis.com/youtube/v3/channels"
+LAST_STATS = Stats(0, 0)
+
+
+async def get_stats(bot: Bot) -> Stats:
+    link = (
+        f"{BASE_URL}?part=statistics&id={CONFIG.mcoding_yt_id}"
+        f"&key={CONFIG.yt_api_key}"
+    )
```
Lunarmagpie, line 62: This URL should be abstracted by a `YouTubeAPI` object.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+    if not response:
+        return LAST_STATS
+
+    items = response.get("items")
+    if not items:
+        return LAST_STATS
+
+    channel = items[0]
+    if channel.get("id") != CONFIG.mcoding_yt_id:
+        return LAST_STATS
```
Lunarmagpie, line 77: This should log a warning imo and why does it need to be mcoding? Any channel should be able to work.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+async def get_stats(bot: Bot) -> Stats:
+    link = (
+        f"{BASE_URL}?part=statistics&id={CONFIG.mcoding_yt_id}"
+        f"&key={CONFIG.yt_api_key}"
+    )
+
+    session = await bot.get_session()
+    async with session.get(link) as res:
+        response: dict[str, Any] = await res.json()
```
Lunarmagpie, line 66: What if `res.status` isn't 200? If this is in a `YouTubeAPI` object it can log a warning and return None.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+if TYPE_CHECKING:
+    from mcodingbot.bot import Bot
+
+
+async def loop_update_channels(bot: Bot) -> None:
+    while True:
+        try:
+            await update_channels(bot)
+        except Exception:
+            print("Failed to update channel stats:")
```
Lunarmagpie, line 20: use the logging module. I dont think this should be here anyway tbh. Errors should be handled in a way thats more abstracted and more detailed. (use a `YouTubAPI` class because that will deal with the problem).

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+    if not subs or not views:
+        return LAST_STATS
+
+    LAST_STATS.subs = subs
+    LAST_STATS.views = views
+    return LAST_STATS
+
+
+def display_stats(stat: int | float) -> str:
+    int_stat = stat
```
Lunarmagpie, line 95: this is useless

Lunarmagpie (changes requested):

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+        return LAST_STATS
+
+    statistics = channel.get("statistics")
+    if not statistics:
+        return LAST_STATS
+
+    subs = float(statistics.get("subscriberCount", 0))
+    views = float(statistics.get("viewCount", 0))
+
+    if not subs or not views:
```
trag1c, line 86:
```suggestion
    if not (sub and views):
```

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+        unit = "M"
+
+    pretty_stat = round(pretty_stat, 2)
+
+    exp_stat = round(log(int_stat, 2), 1)
+    # ^ this might not be as accurate as the member count thing when
+    # someone picky actually calculates it, but I suppose it's not
+    # gonna be such a problem if it's gonna be shown as e.g. "44.3K"
+
+    if exp_stat % 1 == 0:
```
trag1c, line 114:
```suggestion
    if exp_stat.is_integer():
```

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+        exp_stat = int(exp_stat)
+
+    if pretty_stat % 1 == 0:
```
trag1c, line 117:
```suggestion
    if pretty_stat.is_integer():
```

mcodingbot/tasks/stats.py
trag1c: it's rounding to 1 decimal place 👨‍🦯👨‍🦯👨‍🦯

mcodingbot/tasks/stats.py
trag1c: rounding to 2 decimal places 👨‍🦯👨‍🦯👨‍🦯

trag1c (changes requested):
I agree with Lunar in most cases 👍

mcodingbot/tasks/stats.py
CircuitSacul: I'm not willing to re-figure out how to use youtube's garbage api or their thingy. This code is just copy+pasted from mCodingBot-pincer with a few modifications

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,129 @@
...
+
+    channel = items[0]
+    if channel.get("id") != CONFIG.mcoding_yt_id:
+        LOGGER.error("Channel ID was not mCoding.")
+        return _last_known_stats
+
+    statistics = channel.get("statistics")
+    if not statistics:
+        LOGGER.error("mCoding channel did not contain 'statistics'.")
+        return _last_known_stats
```
trag1c, line 88: I don't like the repeated return here but atm I have no idea how to improve this

trag1c (approved):

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+    sub_channel = bot.cache.get_guild_channel(CONFIG.sub_count_channel)
+    view_channel = bot.cache.get_guild_channel(CONFIG.view_count_channel)
+    member_channel = bot.cache.get_guild_channel(CONFIG.member_count_channel)
+
+    assert sub_channel
+    assert view_channel
+    assert member_channel
+
+    await sub_channel.edit(name=f"Subs: {display_stats(stats.subs)}")
+    await view_channel.edit(name=f"Views: {display_stats(stats.views)}")
```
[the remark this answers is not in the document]
CircuitSacul: That feels unecessary - also hikari seems to have a bug with ratelimiting when you gather tasks.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,128 @@
...
+
+    guild = bot.cache.get_guild(CONFIG.mcoding_server)
+    assert guild
+    member_count = guild.member_count
+    assert member_count is not None
+
+    await member_channel.edit(name=f"Members: {display_stats(member_count)}")
```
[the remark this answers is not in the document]
CircuitSacul: Fixed

mcodingbot/bot.py
Lunarmagpie: If its gonna be left here it shouldn't be protected.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,120 @@
...
+
+async def loop_update_channels(bot: Bot) -> None:
+    while True:
+        try:
+            await update_channels(bot)
+        except Exception:
+            print("Failed to update channel stats:")
+            traceback.print_exc()
+
+        await asyncio.sleep(5 * 60)
```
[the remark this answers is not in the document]
Lunarmagpie: Imagine implementing a useless task system. Couldn't be me. Gonna fix this once the next crescent release comes out.

mcodingbot/tasks/stats.py
Lunarmagpie: Im not happy with this code quality. I'm gonna make a PR to fix this later then.

CircuitSacul: I made it use LOGGER, so that's better.

mcodingbot/bot.py
Lunarmagpie: Its better to use a session property and set `self._session` on `hikari.StartedEvent` or in the start method because thats already overridden imo

Lunarmagpie: 1 - Then just throw it in the `start` method.
2 - Thats not possible.
3 - Thats not possible anyway because you need an async context to do that. Everything async always starts after started event.

Do what you want though. Im not spending more time bikeshedding XD

mcodingbot/tasks/stats.py
CircuitSacul: Feels overkill. If we had like 10 if-else it would be worth it, but for just 3 I think this is fine.

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,139 @@
...
+async def update_channels(bot: Bot) -> None:
+    stats = await get_stats(bot)
+
+    sub_channel = bot.cache.get_guild_channel(CONFIG.sub_count_channel)
+    view_channel = bot.cache.get_guild_channel(CONFIG.view_count_channel)
+    member_channel = bot.cache.get_guild_channel(CONFIG.member_count_channel)
+
+    assert sub_channel
+    assert view_channel
+    assert member_channel
```
Sigmanificient, line 37: i'm not a fan of having `assert` elsewhere than within unittests tbh
also since theses lines are basically the same repeated 3 times, why not extracting it to a method?

CircuitSacul: Removed the asserts, they were no longer needed.

Lunarmagpie (approved):

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+async def loop_update_channels(bot: Bot) -> None:
+    while True:
+        try:
+            await update_channels(bot)
+        except Exception:
+            LOGGER.error("Failed to update channel stats:", exc_info=True)
+
+        await asyncio.sleep(5 * 60)
```
Sigmanificient, line 25: shouldn't this be updated too since `update_channels` no longer throw assertion errors?

mcodingbot/tasks/stats.py
```diff
@@ -0,0 +1,136 @@
...
+            "Statistics did not contain subscriberCount and viewCount."
+        )
+        return _last_known_stats
+
+    _last_known_stats.subs = subs
+    _last_known_stats.views = views
+    return _last_known_stats
+
+
+def display_stats(stat: int | float) -> str:
```
Sigmanificient, line 112: So we're using py 3.10+ annotations?

mcodingbot/tasks/stats.py
CircuitSacul: It will still raise errors if it's missing permissions to edit channels, or if discord's api is down, etc.

mcodingbot/tasks/stats.py
CircuitSacul: We have a `from __future__ import annotations` statement, so 3.9+ annotations work in older python versions.

Sigmanificient: oh i didn't knew that, how much is it backwards compatible?

CircuitSacul: min supported version is py 3.8

mcodingbot/tasks/stats.py
Sigmanificient: oh i see

Sigmanificient (approved):

PR #20, opened 2022-07-21
Update crescent to 0.2.4

Lunarmagpie:
This PR also makes updating youtube stuff into a crescent task. A new `Plugin` type is also added where `Plugin.app` is the mcoding bot type for type safety. All the plugins now use this type for consistency.
I also switched loading plugins to use `load_folder` because its a cleaner way to load plugins released in 0.2.4.
Plugin names are also removed because those are deprecated.

### DIFF
--- a/mcodingbot/utils/plugin.py
+++ b/mcodingbot/utils/plugin.py
@@ -13,6 +13,4 @@ __all__: typing.Sequence[str] = ("Plugin",)
 class Plugin(crescent.Plugin):
     @property
     def app(self) -> Bot:
-        # NOTE: Worried that not using `if TYPE_CHECKING` for `Bot` will cause
-        # circular imports later down the line.
         return typing.cast("Bot", super().app)
--- a/mcodingbot/utils/plugin.py
+++ b/mcodingbot/utils/plugin.py
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
-import crescent
 import typing
+
+import crescent
 
 if typing.TYPE_CHECKING:
     from mcodingbot.bot import Bot
--- a/mcodingbot/utils/plugin.py
+++ b/mcodingbot/utils/plugin.py
@@ -1,0 +1,13 @@
+import crescent
+import typing
+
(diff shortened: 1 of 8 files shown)


mcodingbot/utils/plugin.py
```diff
@@ -0,0 +1,18 @@
...
+    from mcodingbot.bot import Bot
+
+__all__: typing.Sequence[str] = ("Plugin",)
+
+
+class Plugin(crescent.Plugin):
+    @property
+    def app(self) -> Bot:
+        # NOTE: Worried that not using `if TYPE_CHECKING` for `Bot` will cause
+        # circular imports later down the line.
```
CircuitSacul, line 17: Since it's only used as a type, using `if typing.TYPE_CHECKING` makes sense even if there's no risk of circular imports.

Lunarmagpie: Do you want me to remove the comment?

CircuitSacul: I don't see any reason to keep it, personally.

mcodingbot/utils/plugin.py
```diff
@@ -0,0 +1,16 @@
...
+if typing.TYPE_CHECKING:
+    from mcodingbot.bot import Bot
+
+__all__: typing.Sequence[str] = ("Plugin",)
+
+
+class Plugin(crescent.Plugin):
+    @property
+    def app(self) -> Bot:
+        return typing.cast("Bot", super().app)
```
Jonxslays, line 16: Why is this using `super()` and not self?

Lunarmagpie: pretty sure `self.app` would cause infinite recursion

Jonxslays: oh, recursion - i forgot when i did this i used bot and app properties to get around that.

Lunarmagpie:
Bump

Endercheif:
@CircuitSacul

PR #28, opened 2022-08-24
fix the member counter

CircuitSacul:
This will make it so that the member count channel will actually update over time.

### DIFF
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -55,7 +55,7 @@ async def update_channels(bot: Bot) -> None:
         return LOGGER.warning(
             "Couldn't find mCoding guild, not updating member count."
         )
-    guild_aprox_members = guild.member_count
+    guild_approx_members = guild.member_count
     if guild_aprox_members is None:
         return LOGGER.warning("Cached guild has no aproximate member count.")
     cached_members = len(bot.cache.get_members_view_for_guild(guild.id))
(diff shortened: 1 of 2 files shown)


Endercheif (approved):
👍

qexat (approved):
Looks fine to me

mcodingbot/tasks/stats.py
```diff
@@ -46,16 +46,25 @@ async def update_channels(bot: Bot) -> None:
...
     if not guild:
         return LOGGER.warning(
             "Couldn't find mCoding guild, not updating member count."
         )
-    if not member_channel:
-        return LOGGER.warning("No member count channel to update stats for.")
-    member_count = guild.member_count
-    if member_count is None:
-        return LOGGER.error("Cached guild didn't contain member count.")
+    guild_aprox_members = guild.member_count
```
Lunarmagpie, line 58: I don't think the `guild_approx_member` variable increases clarity

Lunarmagpie (changes requested):

mcodingbot/tasks/stats.py
CircuitSacul: It has to be stored so I can check if it's None, but still make mypy happy.

qexat: There is also a little typo in the name.

Lunarmagpie: Mypy moment

Lunarmagpie, line 58:
```suggestion
    guild_approx_members = guild.member_count
```

Lunarmagpie (changes requested):

Lunarmagpie (approved):

qexat:
:rocket:

PR #29, opened 2022-08-24
feat: RUST_REGEX supports more messages

qexat:
:rocket: :fire: :stars:

### DIFF
--- a/mcodingbot/plugins/reactions.py
+++ b/mcodingbot/plugins/reactions.py
@@ -5,7 +5,7 @@ import hikari
 
 from mcodingbot.utils import Plugin
 
-RUST_REGEX = re.compile(r"\brust\b", flags=re.I)
+RUST_REGEX = re.compile(r"(\brust\b|\bblazingly fast\b)", flags=re.I)
 
 plugin = Plugin()
 
--- a/mcodingbot/plugins/reactions.py
+++ b/mcodingbot/plugins/reactions.py
@@ -5,7 +5,7 @@ import hikari
 
 from mcodingbot.utils import Plugin
 
-RUST_REGEX = re.compile(r"(\brust\b|\bblazingly fast\b)", flags=re.I)
+RUST_REGEX = re.compile(r"(\brust\b|\bblazingly\s+fast\b)", flags=re.I)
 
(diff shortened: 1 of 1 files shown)


Lunarmagpie (changes requested):
Should use a word boundary instead of a space for `blazingly fast` imo

qexat:
> Should use a word boundary instead of a space for `blazingly fast` imo

how so? it is already using `\b`...?

qexat:
do you mean `\s`?

Lunarmagpie:
> 
> how so? it is already using `\b`.

Like `\bblazingly\bfast\b`

CircuitSacul:
> how so? it is already using \b...?

I think lunar means in between the words (e.x blazingly\bfast)

qexat:
> > how so? it is already using `\b`.
> 
> Like `\bblazingly\bfast\b`

that's kinda weird to me tbh, like any word boundary between blazingly and fast? :thinking:

qexat:
i'll change it anyway

Lunarmagpie (approved):

MithicSpirit:
Pretty sure that `\b` doesn't actually match any characters and only boundaries between characters, so now this doesn't match anything.

MithicSpirit:
imo undo that change. just leave it as a space. If you really want to match different types of spaces you could match on space, nbsp, zwsp, tab, newline, underscore, hyphen, etc.

mcodingbot/plugins/reactions.py
```diff
@@ -5,7 +5,7 @@
 
 from mcodingbot.utils import Plugin
 
-RUST_REGEX = re.compile(r"\brust\b", flags=re.I)
+RUST_REGEX = re.compile(r"(\brust\b|\bblazingly fast\b)", flags=re.I)
```
trag1c, line 8:
```suggestion
RUST_REGEX = re.compile(r"(\brust\b|\bblazingly\s+fast\b)", flags=re.I)
```

qexat:
@Lunarmagpie @CircuitSacul can one of you approve to merge please?

trag1c (approved):

PR #31, opened 2022-09-09
Minor improvements; changed stat format

trag1c:
### Added
- Helper functions for `display_stats`

### Changed
- Made the `Config.load` class method actually use `cls` over `Config`
- Made `Config.save` use `type(self)` over `self.__class__`
- Made `typing` imports and typehints consistent
- Simplified `json.loads(f.read())` to `json.load(f)`
- Simplified `f.write(json.dumps(obj))` to `json.dump(obj, f)`
- Simplified `math.log(n, 2)` to `math.log2(n)`
- Changed exponents to literals in `display_stats`
- `display_stats`' exponents and approximate values will now be floored instead of rounded
  > Example for 1007:
  > `2**10 (1.01K)` -> `2**9.9 (1K)` (`2**10` would actually be displayed once 1024 is hit)
- `display_stats` will now display 2 decimal places instead of 1 for exponents
  > Example for 1000:
  > `2**10 (1K)` -> `2**9.96 (1K)`

### Removed
- Deleted a random comment that did not seem to fit anywhere (?)

### DIFF
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
+from decimal import Decimal
 from math import log2
 from typing import TYPE_CHECKING, Any
 
@@ -121,6 +122,18 @@ async def get_stats(bot: Bot) -> Stats:
     return _last_known_stats
 
 
+def format_float(n: int | float) -> int | float:
+    if float(n).is_integer():
+        return int(n)
+    return n
+
+
+def floor(number: int | float, ndigits: int = 0) -> float:
+    n: int | float = 10 ** ndigits
+    return int(number * n) / n
+
+
+
 def display_stats(stat: int | float) -> str:
     if stat < 1_000:
         pretty_stat = stat
@@ -132,16 +145,10 @@ def display_stats(stat: int | float) -> str:
         pretty_stat = stat / 1_000_000
         unit = "M"
 
-    pretty_stat = int(pretty_stat * 100) / 100
-    exp_stat = int(log2(stat) * 10) / 10
+    pretty_stat = format_float(floor(pretty_stat, 2))
+    exp_stat = format_float(floor(log2(stat), 2))
     # ^ this might not be as accurate as the member count thing when
     # someone picky actually calculates it, but I suppose it's not
     # gonna be such a problem if it's gonna be shown as e.g. "44.3K"
-
-    if float(exp_stat).is_integer():
-        exp_stat = int(exp_stat)
-
-    if float(pretty_stat).is_integer():
-        pretty_stat = int(pretty_stat)
 
     return f"2**{exp_stat} ({pretty_stat}{unit})"
--- a/mcodingbot/tasks/stats.py
+++ b/mcodingbot/tasks/stats.py
@@ -121,7 +121,7 @@ async def get_stats(bot: Bot) -> Stats:
     return _last_known_stats
 
 
-def format_float(n: int | float) -> int | float:
+def strip_trailing_zeros(n: int | float) -> int | float:
     if float(n).is_integer():
         return int(n)
     return n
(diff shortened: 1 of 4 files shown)


mcodingbot/tasks/stats.py
```diff
@@ -122,27 +121,32 @@ async def get_stats(bot: Bot) -> Stats:
...
+
+
+def floor(number: int | float, ndigits: int = 0) -> float:
+    n: int | float = 10**ndigits
+    return int(number * n) / n
+
+
 def display_stats(stat: int | float) -> str:
-    if stat < 10**3:
+    if stat < 1_000:
```
Lunarmagpie, line 136: Imo 10**N was more readable here

mcodingbot/utils/plugin.py
```diff
@@ -1,16 +1,16 @@
 from __future__ import annotations
 
-import typing
+from typing import cast, Sequence, TYPE_CHECKING
```
Lunarmagpie, line 3: Please revert.

mcodingbot/tasks/stats.py
```diff
@@ -122,27 +121,32 @@ async def get_stats(bot: Bot) -> Stats:
     return _last_known_stats
 
 
+def format_float(n: int | float) -> int | float:
+    if float(n).is_integer():
+        return int(n)
+    return n
+
+
+def floor(number: int | float, ndigits: int = 0) -> float:
```
Lunarmagpie, line 130: This function does not floor. `truncate_decimals` is closer.

mcodingbot/utils/__init__.py
```diff
@@ -1,5 +1,5 @@
-import typing
+from typing import Sequence
```
Lunarmagpie, line 1: I dont think there should be a from import here

Lunarmagpie (changes requested):
Mcoding bot uses `import typing` primarily I believe

mcodingbot/utils/plugin.py
trag1c: Why?

mcodingbot/utils/__init__.py
trag1c: Why?

mcodingbot/tasks/stats.py
trag1c: Good point, thanks

mcodingbot/utils/__init__.py
Lunarmagpie: I thought this project did `import X`

trag1c:
> Mcoding bot uses `import typing` primarily I believe

Before my PR it was 3x `from typing import ...` vs 2x `import typing`

mcodingbot/tasks/stats.py
```diff
@@ -122,27 +121,32 @@ async def get_stats(bot: Bot) -> Stats:
     return _last_known_stats
 
 
+def format_float(n: int | float) -> int | float:
```
Lunarmagpie, line 124: Would rename this to `strip_trailing_zeros` tbh

Lunarmagpie (changes requested):

mcodingbot/tasks/stats.py
trag1c: Yeah sounds more accurate, thanks!

mcodingbot/tasks/stats.py
trag1c: I guess that's subjective; also `1e3` is another option :trollface:

Lunarmagpie: I think the relationship between 10**N and 10**N-3 shows the intent more clearly than a different number of 0s

CircuitSacul:
You need to run `isort` locally

CircuitSacul (approved):
lgtm, I still want to test it but I can't right now

Lunarmagpie (approved):
its not like 0 or `10**N` matters more than this being merged within the next 50 years

PR #33, opened 2022-09-09
Add enterprise joke command

Lunarmagpie:
It is enterprise. The code is lightly commented so it will be understandable later.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.linting.pylintEnabled": false,
+    "python.linting.mypyEnabled": true,
+    "python.linting.enabled": true
+}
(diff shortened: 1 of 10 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "python.linting.pylintEnabled": false,
+    "python.linting.mypyEnabled": true,
+    "python.linting.enabled": true
+}
```
CircuitSacul, line 5: this should probably be gitignored, but for now could you just delete the file from this branch

CircuitSacul (changes requested):
This feels super overkill. I'm not sure why you made it so complex, but wouldn't it be better to just have a dictionary? In fact, the jokes should really be a .json file that is parsed at startup.


Also not sure why poetry.lock changed?

.vscode/settings.json
trag1c: > This feels super overkill. I'm not sure why you made it so complex

"Enterprise" code

PR #35, opened 2022-09-09
Replaced "if" statements with functions

FuexFollets:


qexat:
A ReactionManager is feasible, but not realistic for a project of this size. I think that simple ifs/elses suffice for this bot.

PR #36, opened 2022-09-09
handle donor role

### LINKED ISSUE #27: Retroactively add Donor role to every user that has the Patron role
Everyone with the Patron role should receive the Donor role, and the bot does somehow do that, but it seems like the bot needs to be online when the user gets the Patron role, so if the bot is offline and a user gets Patron, the user will not get Donor, like, ever.

My suggestion is for the bot to, on a programmed schedule (every day should be more than enough), look at everyone on Patron and make sure they're also on Donor.

CircuitSacul:
closes #27

### DIFF
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import logging
 from typing import Any
 
 import aiohttp
@@ -9,8 +8,6 @@ import hikari
 
 from mcodingbot.config import CONFIG
 from mcodingbot.database.database import Database
-
-LOGGER = logging.getLogger(__name__)
 
 
 class Bot(crescent.Bot):
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -42,20 +42,7 @@ class Bot(crescent.Bot):
 
     async def start(self, *args: Any, **kwargs: Any) -> None:
         self._session = aiohttp.ClientSession()
-
-        self._db = Database("mcodingbot/database/migrations")
-        await self.db.connect(
-            host="localhost",
-            database="mcodingbot",
-            user="mcodingbot",
-            password=CONFIG.db_password,
-        )
-        if self.db.must_create_migrations():
-            LOGGER.info("Creating migrations...")
-            self.db.create_migrations()
-        if await self.db.must_apply_migrations():
-            LOGGER.info("Applying migrations...")
-            await self.db.apply_migrations()
+        self._db = await Database.create()
 
         await super().start(*args, **kwargs)
 
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 import aiohttp
@@ -7,6 +8,9 @@ import crescent
 import hikari
 
 from mcodingbot.config import CONFIG
+from mcodingbot.database.database import Database
+
+LOGGER = logging.getLogger(__name__)
 
 
 class Bot(crescent.Bot):
@@ -22,6 +26,7 @@ class Bot(crescent.Bot):
         self.plugins.load_folder("mcodingbot.tasks")
 
         self._session: aiohttp.ClientSession | None = None
+        self._db: Database | None = None
 
     @property
     def session(self) -> aiohttp.ClientSession:
@@ -29,8 +34,29 @@ class Bot(crescent.Bot):
             raise AttributeError("Session has not been set yet.")
         return self._session
 
+    @property
+    def db(self) -> Database:
+        if not self._db:
+            raise AttributeError("Database has not been set yet.")
+        return self._db
+
     async def start(self, *args: Any, **kwargs: Any) -> None:
         self._session = aiohttp.ClientSession()
+
+        self._db = Database("mcodingbot/database/migrations")
+        await self.db.connect(
+            host="localhost",
+            database="mcodingbot",
+            user="mcodingbot",
+            password=CONFIG.db_password,
+        )
+        if self.db.must_create_migrations():
+            LOGGER.info("Creating migrations...")
+            self.db.create_migrations()
+        if await self.db.must_apply_migrations():
+            LOGGER.info("Applying migrations...")
+            await self.db.apply_migrations()
+
         await super().start(*args, **kwargs)
 
     async def join(self, *args: Any, **kwargs: Any) -> None:
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -4,11 +4,13 @@ from dataclasses import asdict, dataclass
 from pathlib import Path
 from typing import Any, Dict, cast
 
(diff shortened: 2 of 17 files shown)


CircuitSacul:
I finally snuck apgorm back in

Lunarmagpie:
> I finally snuck apgorm back in

LETS FUCKING GO

mcodingbot/bot.py
```diff
@@ -22,15 +26,37 @@ def __init__(self) -> None:
...
             raise AttributeError("Session has not been set yet.")
         return self._session
 
+    @property
+    def db(self) -> Database:
+        if not self._db:
+            raise AttributeError("Database has not been set yet.")
+        return self._db
+
     async def start(self, *args: Any, **kwargs: Any) -> None:
```
Lunarmagpie, line 43: Is is better to put this stuff in a `create_db` function that's subscribed to `hikari.StartingEvent` imo.

mcodingbot/database/converters.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+from decimal import Decimal
+
+from apgorm import Converter
+
+
+class NumericC(Converter[Decimal, int]):
```
Lunarmagpie, line 8: What is C? This class needs a better name that makes it ambitious. It probably stands for Converter but C is used so much to refer to the language (like `c_int`) that's likely what people's first thought will be

mcodingbot/database/models/user.py
```diff
@@ -0,0 +1,21 @@
...
+
+from apgorm import Model, types
+from asyncpg.exceptions import UniqueViolationError
+
+from mcodingbot.database.converters import NumericC
+
+
+class User(Model):
+    user_id = types.Numeric().field().with_converter(NumericC)
+
```
Lunarmagpie, line 11: Don't think we need this whitespace

mcodingbot/plugins/donors.py
```diff
@@ -0,0 +1,95 @@
...
+        await _update_donor_role(self.member.id, self.is_donor)
+
+        await ctx.respond(
+            f"{self.member.mention} {'' if self.is_donor else 'un'}marked"
+            " as a donor.",
+            ephemeral=True,
+        )
+
+
+# if a member receives the patron role (from the Patreon bot), also
```
Lunarmagpie, line 40: You can use a docstring.

mcodingbot/plugins/donors.py
```diff
@@ -0,0 +1,95 @@
...
+# give them the donor role.
+@plugin.include
+@crescent.event
+async def on_member_update(event: hikari.MemberUpdateEvent) -> None:
+    await _give_donor_role_if_donor(event.member)
+
+
+# loop over members and give them the donor role if they deserve it
```
Lunarmagpie, line 48: I don't think this comment is needed

mcodingbot/plugins/donors.py
```diff
@@ -0,0 +1,95 @@
...
+@plugin.include
+@tasks.loop(hours=1)
```
Lunarmagpie, line 50: `tasks.cronjob(0 * * * *)` :)

mcodingbot/plugins/donors.py
```diff
@@ -0,0 +1,95 @@
...
+async def add_donor_role() -> None:
+    for member in plugin.app.cache.get_members_view_for_guild(
```
Lunarmagpie, line 52: Use asyncio.gather

Lunarmagpie (changes requested):

mcodingbot/bot.py
CircuitSacul: Why would we want to create the database after the bot's connected to discord? Events can sometimes trigger before that anyways, it's best to make the DB right away. I could see moving this logic to a method on the Database itself though.

Lunarmagpie: `StartingEvent` fires before you connect to discord

mcodingbot/plugins/donors.py
CircuitSacul: A loop works perfectly fine, using a cronjob seems pointless

Lunarmagpie: :(

mcodingbot/bot.py
CircuitSacul: Okay but subscribing to an event is pointless if you can do it before connecting. It just adds complexity to the bot.

mcodingbot/database/converters.py
CircuitSacul: Should I use NumericInt, DecimalInt, NumericConverter, or DecimalConverter?

Lunarmagpie: Probably NumericConverter

mcodingbot/bot.py
Lunarmagpie: I feel like moving the logic to the `Database` class then subscribing the connect function would be the simplest

CircuitSacul: There's no reason to subscribe to an event. Whatever code we call when the StartingEvent is fired can just be put in the start method.

CircuitSacul: I moved the logic to the Database class though.

Lunarmagpie: I think that's good enough

Lunarmagpie (approved):

Lunarmagpie:
Just double check the whitespace spot

mcodingbot/config.py
```diff
@@ -6,18 +6,22 @@
 from pathlib import Path
 from typing import Any, cast
 
-_ALWAYS_SAVE = ["discord_token"]
+_ALWAYS_SAVE = ["discord_token", "db_password"]
```
trag1c, line 9: I guess this could be either a tuple or a set

mcodingbot/database/database.py
```diff
@@ -0,0 +1,32 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Database(apgorm.Database):
+    users = User
+
+    @staticmethod
+    async def create() -> Database:
+        db = Database("mcodingbot/database/migrations")
```
trag1c, line 18: Why not class method if it's returning the instance of itself?

CircuitSacul: Because then I would have to write
```py
@classmethod
async def create(cls) -> ???:
    ...
```

Either I need a TypeVar, or I need to specific the return type is Database. And if I'm already using Database explicitly, it felt more consistent to use that too.

If you still think it should be a classmethod I'll change it.

Lunarmagpie: I think a classmethod is better

Endercheif: I agree with trag1c and lunar here. It makes more sense and is more clear.

Endercheif (approved):

Lunarmagpie (approved):

trag1c (approved):

PR #43, opened 2022-09-12
optionally show embeds for /pep

### LINKED ISSUE #40: Embed reply for PEP detection
- [ ]  Add an option to allow the default link embed in the `/pep` command
- [ ] During detection for a message with pep in it, instead of sending links, send one embed listing all PEPs that were mentioned and their respective links.

> nah, but I was thinking about making the entire reply an embed so you can have hyperlinks when multiple PEPs are mentioned maybe

_Originally posted by @trag1c in https://github.com/mcb-dev/mCodingBot/pull/39#discussion_r968876667_

CircuitSacul:
closes #40

### DIFF
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -10,7 +10,7 @@ plugin = Plugin()
 PEP_REGEX = re.compile(r"pep *(?P<pep>\d{1,4})", re.IGNORECASE)
 
 
-def get_pep_link(pep_number: int, hide_embed: bool) -> str:
+def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
     url = f"https://peps.python.org/pep-{pep_number:04}/"
     return f"<{url}>" if hide_embed else url
 
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -10,8 +10,9 @@ plugin = Plugin()
 PEP_REGEX = re.compile(r"pep *(?P<pep>\d{1,4})", re.IGNORECASE)
 
 
-def get_pep_link(pep_number: int) -> str:
-    return f"<https://peps.python.org/pep-{pep_number:04}/>"
+def get_pep_link(pep_number: int, hide_embed: bool) -> str:
+    url = f"https://peps.python.org/pep-{pep_number:04}/"
+    return f"<{url}>" if hide_embed else url
 
 
 @plugin.include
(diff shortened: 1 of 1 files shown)


mcodingbot/plugins/python.py
```diff
@@ -10,8 +10,9 @@
 PEP_REGEX = re.compile(r"pep *(?P<pep>\d{1,4})", re.IGNORECASE)
 
 
-def get_pep_link(pep_number: int) -> str:
-    return f"<https://peps.python.org/pep-{pep_number:04}/>"
+def get_pep_link(pep_number: int, hide_embed: bool) -> str:
```
Lunarmagpie, line 13: I prefer this signature for functions like this
```suggestion
def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
```

Lunarmagpie (changes requested):

mcodingbot/plugins/python.py
```diff
@@ -22,9 +23,14 @@ class PEPCommand:
     pep_number = crescent.option(
         int, "The PEP number.", name="pep-number", min_value=0, max_value=9999
     )
+    hide_embed = crescent.option(
```
qexat, line 26: Imo, it should be the opposite (aka `show_embed` defaulting to `True`) but that's up to code style

CircuitSacul: For the command option, I agree. I'll change that.

Lunarmagpie (approved):

mcodingbot/plugins/python.py
qexat: boolean args should most of the time be keyword-only yeah

Endercheif (approved):

mcodingbot/plugins/python.py
```diff
@@ -42,7 +48,7 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         return
 
     pep_links_message = "\n".join(
-        f"PEP {pep_number}: {get_pep_link(pep_number)}"
+        f"PEP {pep_number}: {get_pep_link(pep_number, hide_embed=True)}"
```
qexat, line 51: do you mean, `hide_embed=False`?

qexat: honestly it would be better to just call it `show_embed` as well

CircuitSacul: No I didn't mean that, the reply can already be annoying, even without the embed.

PR #46, opened 2022-09-12
Adjusted displaying PEPs

trag1c:


isFakeAccount:
I approve

mcodingbot/plugins/python.py
```diff
@@ -7,7 +7,7 @@
 
 plugin = Plugin()
 
-PEP_REGEX = re.compile(r"pep *(?P<pep>\d{1,4})", re.IGNORECASE)
+PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d+)", re.IGNORECASE)
```
Lunarmagpie, line 10: why `+` instead of {1-4}?

trag1c: Screen Shot 2022-09-13 at 00 16 41
(pasting here for context)

Lunarmagpie, line 10:
```suggestion
PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}[^\d])", re.IGNORECASE)
```

Lunarmagpie: Throwing a word boundary at the end also works if you dont want to match `pep 1234abc`.

CircuitSacul: trag1c's is better because `pep 9.` should match, but you don't want `pep 99999` to become `pep 9999`

trag1c: This one won't work, requires another character after the PEP number

trag1c: image

mcodingbot/plugins/python.py
```diff
@@ -38,12 +38,17 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
...
         return
 
     pep_links_message = "\n".join(
         f"PEP {pep_number}: {get_pep_link(pep_number)}"
-        for pep_number in pep_refs
+        for pep_number in pep_refs[:5]
     )
 
+    if len(pep_refs) > 5:
+        pep_links_message += f"\n({len(pep_refs) - 5} PEPs omitted)"
```
CircuitSacul, line 52: I see `len(pep_refs)` twice. Maybe
```py
if (num := len(pep_refs)) > 5:
    pep_links_message += f"\n({num-5} PEPs omitted)"
```

mcodingbot/plugins/python.py
```diff
@@ -38,12 +38,17 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         for ref in re.finditer(PEP_REGEX, event.message.content)
     ]
 
+    pep_refs = sorted(set(filter(lambda x: x < 10000, pep_refs)))
```
CircuitSacul, line 41:
```py
pep_refs = sorted({ref for ref in pep_refs if ref < 10000})
```

CircuitSacul (changes requested):

mcodingbot/plugins/python.py
Lunarmagpie: image


Word boundary works

mcodingbot/plugins/python.py
trag1c: that's how I originally did it but `filter` felt more clear to me

mcodingbot/plugins/python.py
Lunarmagpie: > trag1c's is better because `pep 9.` should match, but you don't want `pep 99999` to become `pep 9999`

pep 99999 doesn't match to begin with for mine

mcodingbot/plugins/python.py
trag1c: can do set comp if u want

mcodingbot/plugins/python.py
isFakeAccount: How does `?P` even work? What is it for?

trag1c: > How does `?P` even work? What is it for?

Named capture group; in other syntaxes it's just `?` I believe

CircuitSacul: > > trag1c's is better because `pep 9.` should match, but you don't want `pep 99999` to become `pep 9999`
> 
> pep 99999 doesn't match to begin with for mine

Yes but your method breaks if `pep <x>` isn't surrounded by spaces. We don't want that. trag1c's method works for every case so there's no reason to re-engineer the regex.

CircuitSacul: > How does `?P` even work? What is it for?

https://stackoverflow.com/questions/10059673/named-regular-expression-group-pgroup-nameregexp-what-does-p-stand-for

mcodingbot/plugins/python.py
CircuitSacul: I think set comp is more clear in this situation.

mcodingbot/plugins/python.py
Lunarmagpie: image
Mine fails in this case but tragic's does as well for some reason. Mine does work with trailing spaces.
anyway why does that fail lol

CircuitSacul:
```py
REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
```

Since Lunar's regex works for all the cases I can find, we wouldn't have to check that the regex is < 1000 inside comp.

CircuitSacul (approved):

Lunarmagpie (approved):

PR #47, opened 2022-09-12
Restart commands & auto-pull changes

### LINKED ISSUE #41: Add /gitpull and /restart commands
So I don't have to SSH every time to update the bot.

CircuitSacul:
closes #41

### DIFF
--- a/mcodingbot/plugins/bot_management.py
+++ b/mcodingbot/plugins/bot_management.py
@@ -1,0 +1,22 @@
+from __future__ import annotations
+
+import hikari
+import crescent
+
+from mcodingbot.config import CONFIG
+
+plugin = crescent.Plugin()
+
+
+@plugin.include
+@crescent.command(
+    name="restart",
+    description="Restarts the bot.",
+    default_member_permissions=hikari.Permissions.MODERATE_MEMBERS,
+    dm_enabled=False,
+    guild=CONFIG.mcoding_server,
+)
+class RestartBot:
+    async def callback(self, ctx: crescent.Context) -> None:
+        await ctx.respond("Restarting bot...")
+        await ctx.app.close()
--- a/run.sh
+++ b/run.sh
@@ -2,6 +2,8 @@
 
 while true
 do
+    git pull
+    "${1:-python}" -m poetry install
     "${1:-python}" -m poetry run python -m mcodingbot
 
     echo "Hit CTRL+C to stop..."


mcodingbot/plugins/bot_management.py
```diff
@@ -0,0 +1,22 @@
...
+from mcodingbot.config import CONFIG
+
+plugin = crescent.Plugin()
+
+
+@plugin.include
+@crescent.command(
+    name="restart",
+    description="Restarts the bot.",
+    default_member_permissions=hikari.Permissions.MODERATE_MEMBERS,
```
Lunarmagpie, line 15: I feel like this one makes more sense tbh 
```suggestion
    default_member_permissions=hikari.Permissions.ADMINISTRATOR,
```

mcodingbot/plugins/bot_management.py
```diff
@@ -0,0 +1,22 @@
...
+    dm_enabled=False,
+    guild=CONFIG.mcoding_server,
+)
+class RestartBot:
+    async def callback(self, ctx: crescent.Context) -> None:
+        await ctx.respond("Restarting bot...")
+        await ctx.app.close()
```
Lunarmagpie, line 22:
```suggestion
async def restart_bot(ctx: crescent.Context) -> None:
    await ctx.respond("Restarting bot...")
    await ctx.app.close()
```

Lunarmagpie (changes requested):

mcodingbot/plugins/bot_management.py
CircuitSacul: I did this to be consistent

https://github.com/mcb-dev/mCodingBot/blob/main/mcodingbot/plugins/info.py#L10

We can change this in a seperate pr if we want to.

Lunarmagpie: I feel like that one should be changed. I want to give function commands some love :(
Links would benefit a bit form a lower indentation level as well

CircuitSacul: If it's going to be changed I think it should be done in a new pr, since so far every command (even ones without options) are class commands.

Lunarmagpie: :+1:

Lunarmagpie (approved):

trag1c (approved):

PR #48, opened 2022-09-13
Add Pep validation and name to output

Lunarmagpie:
This is most definitely draft quality code.

image

### DIFF
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -6,6 +6,7 @@ from crescent.ext import tasks
 
 from mcodingbot.utils import PepManager, Plugin
 
+
 class PepPlugin(Plugin):
     def __init__(self) -> None:
         self.peps = PepManager()
@@ -33,6 +34,7 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
         hikari.ButtonStyle.SECONDARY, encode_dismiss_button_id(id)
     ).set_label("Dismiss").add_to_container()
     return action_row
+
 
 @plugin.include
 @tasks.loop(hours=1)
@@ -79,9 +81,7 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         return
 
     peps = (plugin.peps.get(pep_number) for pep_number in pep_refs)
-    pep_links_message = "\n".join(
-        str(pep) for pep in peps if pep
-    )
+    pep_links_message = "\n".join(str(pep) for pep in peps if pep)
 
     if (pep_number := len(pep_refs)) > 5:
         pep_links_message += f"\n({pep_number - 5} PEPs omitted)"
@@ -89,7 +89,7 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
     await event.message.respond(
         pep_links_message,
         component=get_dismiss_button(event.author.id),
-        reply=True
+        reply=True,
     )
 
 
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -8,10 +8,10 @@ from mcodingbot.config import CONFIG
 from mcodingbot.utils import PEPManager, Plugin
 
 plugin = Plugin()
+pep_manager = PEPManager()
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 DISMISS_BUTTON_ID = "dismiss"
-PEP_MANAGER = PEPManager()
 
 
 def encode_dismiss_button_id(id: hikari.Snowflake) -> str:
@@ -33,7 +33,7 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
 @plugin.include
 @tasks.cronjob("@daily")
 async def update_peps() -> None:
-    await PEP_MANAGER.fetch_pep_info(plugin.app)
+    await pep_manager.fetch_pep_info(plugin.app)
 
 
 @plugin.include
@@ -49,7 +49,7 @@ class PEPCommand:
     )
 
     async def callback(self, ctx: crescent.Context) -> None:
-        if not (pep := PEP_MANAGER.get(self.pep_number)):
+        if not (pep := pep_manager.get(self.pep_number)):
             await ctx.respond(
                 f"{self.pep_number} is not a valid PEP.", ephemeral=True
             )
@@ -69,7 +69,7 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         for ref in re.finditer(PEP_REGEX, event.message.content)
     ]
 
-    peps = map(PEP_MANAGER.get, sorted(set(pep_refs))[:5])
+    peps = map(pep_manager.get, sorted(set(pep_refs))[:5])
     pep_links_message = "\n".join(str(pep) for pep in peps if pep)
 
     if not pep_links_message:
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -6,12 +6,12 @@ from crescent.ext import tasks
 
 from mcodingbot.utils import PepManager, Plugin
 
-
 plugin = Plugin()
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 DISMISS_BUTTON_ID = "dismiss"
 PEP_MANAGER = PepManager()
+
 
 def encode_dismiss_button_id(id: hikari.Snowflake) -> str:
     return f"{DISMISS_BUTTON_ID}:{id}"
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -68,13 +68,11 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         for ref in re.finditer(PEP_REGEX, event.message.content)
     ]
 
-    pep_refs = sorted(set(pep_refs))[:5]
+    peps = map(PEP_MANAGER.get, sorted(set(pep_refs))[:5])
+    pep_links_message = "\n".join(str(pep) for pep in peps if pep if pep)
 
-    if not pep_refs:
+    if not pep_links_message:
         return
-
-    peps = (PEP_MANAGER.get(pep_number) for pep_number in pep_refs)
-    pep_links_message = "\n".join(str(pep) for pep in peps if pep)
 
     if (pep_number := len(pep_refs)) > 5:
         pep_links_message += f"\n({pep_number - 5} PEPs omitted)"
--- a/mcodingbot/plugins/python.py
+++ b/mcodingbot/plugins/python.py
@@ -10,8 +10,9 @@ plugin = Plugin()
 PEP_REGEX = re.compile(r"pep *(?P<pep>\d{1,4})", re.IGNORECASE)
 
 
-def get_pep_link(pep_number: int) -> str:
-    return f"<https://peps.python.org/pep-{pep_number:04}/>"
+def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
+    url = f"https://peps.python.org/pep-{pep_number:04}/"
+    return f"<{url}>" if hide_embed else url
 
 
 @plugin.include
@@ -22,9 +23,14 @@ class PEPCommand:
     pep_number = crescent.option(
         int, "The PEP number.", name="pep-number", min_value=0, max_value=9999
     )
+    show_embed = crescent.option(
+        bool, "Whether to show the embed.", name="show-embed", default=True
+    )
 
     async def callback(self, ctx: crescent.Context) -> None:
-        await ctx.respond(get_pep_link(self.pep_number))
+        await ctx.respond(
+            get_pep_link(self.pep_number, hide_embed=not self.show_embed)
+        )
 
 
 @plugin.include
@@ -42,7 +48,7 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         return
 
     pep_links_message = "\n".join(
-        f"PEP {pep_number}: {get_pep_link(pep_number)}"
+        f"PEP {pep_number}: {get_pep_link(pep_number, hide_embed=True)}"
         for pep_number in pep_refs
     )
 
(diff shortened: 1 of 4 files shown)


Lunarmagpie:
I think I will store the PepManager on the plugin

mcodingbot/plugins/python.py
```diff
@@ -2,17 +2,17 @@
...
 plugin = Plugin()
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 
 
-def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
-    url = f"https://peps.python.org/pep-{pep_number:04}/"
-    return f"<{url}>" if hide_embed else url
+def get_pep(pep_number: int) -> Pep | None:
+    return plugin.app.peps.get(pep_number)
```
CircuitSacul, line 15: This feels like a useless function now

mcodingbot/plugins/python.py
```diff
@@ -49,12 +53,19 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
     if not pep_refs:
         return
 
+    peps = (get_pep(pep_number) for pep_number in pep_refs[:5])
+
```
CircuitSacul, line 57: Instead of doing list comp again, just change line 51

```py
pep_refs = sorted(set(pep_refs))[:5]
```

mcodingbot/utils/peps.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, bot: Bot) -> None:
+        self._bot = bot
+        self._peps: dict[int, Any] = {}
+
+    async def fetch_pep_info(self) -> None:
+        async with self._bot.session.get(
+            "https://peps.python.org/api/peps.json"
+        ) as resp:
+            if resp.status != 200:
+                _LOG.warning("Could not fetch peps.")
```
CircuitSacul, line 25: This is an error. Just do `resp.raise_for_status()` though.

CircuitSacul (changes requested):
I would remove the whole thing about embeds/no embeds, since we can now get more useful information. Also if we put the result inside an embed, we can use masked links so the result is less ugly.

mcodingbot/utils/peps.py
```diff
@@ -0,0 +1,54 @@
...
+class Pep:
+    number: int
+    title: str
+    authors: str
+    link: str
+    python_version: str | None
+
+    def stringify(self, *, hide_embed: bool) -> str:
+        link = f"<{self.link}>" if hide_embed else self.link
+        return f"PEP {self.number}: {self.title} ({link})"
```
CircuitSacul, line 54: The `__str__` magic method exists, would it make more sense to use that instead? Usually, I let `repr(<dataclass>)` be the default (shows the code-like format) and I use `str(<dataclass>)` for the frontend format.

This only applies if we do away with the hide_embed kwarg.

CircuitSacul (changes requested):

Lunarmagpie:
> I would remove the whole thing about embeds/no embeds, since we can now get more useful information. Also if we put the result inside an embed, we can use masked links so the result is less ugly.

I think you're right tbh 
Cc @trag1c

trag1c:
> > I would remove the whole thing about embeds/no embeds, since we can now get more useful information. Also if we put the result inside an embed, we can use masked links so the result is less ugly.
> 
> I think you're right tbh Cc @trag1c

:+1:

mcodingbot/plugins/python.py
```diff
@@ -2,13 +2,20 @@
...
+from mcodingbot.utils import PepManager, Plugin
 
-plugin = Plugin()
 
-PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
+class PepPlugin(Plugin):
+    def __init__(self) -> None:
+        self.peps = PepManager()
+        super().__init__()
+
```
CircuitSacul, line 14: just make PepManager a constant. No need to subclass the plugin

mcodingbot/plugins/python.py
```diff
@@ -63,23 +75,21 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
         for ref in re.finditer(PEP_REGEX, event.message.content)
     ]
 
-    pep_refs = sorted(set(pep_refs))
+    pep_refs = sorted(set(pep_refs))[:5]
```
CircuitSacul, line 78:
```py
pep_refs = map(plugin.peps.get, sorted(set(pep_refs))[:5])
```
maybe?

CircuitSacul (changes requested):

trag1c (changes requested):
Rename `Pep` and `PepManager` to `PEP` and `PEPManager`

mcodingbot/utils/peps.py
```diff
@@ -0,0 +1,53 @@
...
+    async def fetch_pep_info(self, bot: Bot) -> None:
+        async with bot.session.get(
+            "https://peps.python.org/api/peps.json"
+        ) as resp:
+            try:
+                resp.raise_for_status()
+                self._peps = {
+                    int(key): value
+                    for key, value in (await resp.json()).items()
+                }
```
CircuitSacul, line 30: these lines should be put in an `else` clause after the `except` clause.

mcodingbot/utils/peps.py
```diff
@@ -0,0 +1,53 @@
...
+    from mcodingbot.bot import Bot
+
+_LOG = getLogger(__name__)
+
+__all__: Sequence[str] = ("PepManager", "Pep")
+
+
+class PepManager:
+    def __init__(self) -> None:
+        self._peps: dict[int, Any] = {}
```
CircuitSacul, line 19: I think we should store the peps as `Pep` instead of raw dictionaries.

CircuitSacul (changes requested):
I also agree with trag1c's comment

Pep -> PEP
PepManager -> PEPManager

trag1c:
> I also agree with trag1c's comment
> 
> Pep -> PEP PepManager -> PEPManager

You don't actually have to agree since `Pep` and `PepManager` don't follow PEP8 lol

CircuitSacul:
Also, since we now have access to more detailed information, I think that all responses should be embedded. The `Pep` class could have two methods:
`__str__`: Returns the one-line information, using a masked link (instead of putting the links inside parenthesis)
`embed`: Returns a detailed embed of the Pep

Then auto PEP replies would return a single embed containing a line for each individual pep, and the command can return the detailed embed.

CircuitSacul:
> You don't actually have to agree since Pep and PepManager don't follow PEP8 lol

In that case ig they should be left as-is probably.

trag1c:
> > You don't actually have to agree since Pep and PepManager don't follow PEP8 lol
> 
> In that case ig they should be left as-is probably.

what? Pep and PepManager **DON'T** follow PEP8

CircuitSacul:
> what? Pep and PepManager DON'T follow PEP8

Then why did you say I didn't have to agree lol. Go with whatever follows pep8

trag1c:
> > what? Pep and PepManager DON'T follow PEP8
> 
> Then why did you say I didn't have to agree lol. Go with whatever follows pep8

because it doesn't matter whether you agree or not, renaming it is sort of a requirement in this case

Lunarmagpie:
I agree with @trag1c

Lunarmagpie:
> Also, since we now have access to more detailed information, I think that all responses ...

I am against this because of how bulky embeds are. The button takes up enough space already.

Lunarmagpie:
> I agree with @trag1c

Changed my mind. The `Pep` class looks like constant when its capital and I dont like that tbh.

trag1c:
> > I agree with @trag1c
> 
> Changed my mind. The `Pep` class looks like constant when its capital and I dont like that tbh.

Dude that's like naming your functions with camelCase???

Endercheif (changes requested):

mcodingbot/plugins/python.py
```diff
@@ -63,23 +68,19 @@ async def on_message(event: hikari.MessageCreateEvent) -> None:
...
         return
 
-    pep_links_message = "\n".join(
-        f"PEP {pep_number}: {get_pep_link(pep_number, hide_embed=True)}"
-        for pep_number in pep_refs[:5]
-    )
-
-    if (peps := len(pep_refs)) > 5:
-        pep_links_message += f"\n({peps - 5} PEPs omitted)"
+    if (pep_number := len(pep_refs)) > 5:
```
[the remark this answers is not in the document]
Lunarmagpie: renamed to `pep_count`

Endercheif:
For the class renaming discussion,

instead of `Pep` or `PEP` what about something like `PEPObject` or `PEPItem`?

Lunarmagpie:
> Dude that's like naming your functions with camelCase???

I don't understand what you want me to do

trag1c:
> > Dude that's like naming your functions with camelCase???
> 
> I don't understand what you want me to do


"Rename `Pep` and `PepManager` to `PEP` and `PEPManager`"

trag1c:
> > > Dude that's like naming your functions with camelCase???
> > 
> > 
> > I don't understand what you want me to do
> 
> "Rename `Pep` and `PepManager` to `PEP` and `PEPManager`"

If you don't like `PEP` looking like a constant then pick a different name as suggested by @Endercheif

CircuitSacul:
> > Also, since we now have access to more detailed information, I think that all responses ...
> 
> I am against this because of how bulky embeds are. The button takes up enough space already.

Screen Shot 2022-09-13 at 7 07 45 PM

I think it looks great.

Lunarmagpie:
> I think it looks great.

yeah its better than i though actually

CircuitSacul:
> yeah its better than i though actually

Should I just push my changes

Lunarmagpie:
yes

mcodingbot/plugins/python.py
```diff
@@ -2,15 +2,16 @@
...
+from mcodingbot.config import CONFIG
+from mcodingbot.utils import PepManager, Plugin
 
 plugin = Plugin()
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
-
-
 DISMISS_BUTTON_ID = "dismiss"
+PEP_MANAGER = PepManager()
```
CircuitSacul, line 14: so the question is, should this actually be a constant, or just a global?

Usually constants are constant, but this pep manager is modifiable (and it is modified, every hour). Should this maybe be all lower case, and just go under plugin?

CircuitSacul (changes requested):
Also, `Pep->PEPInfo` and `PepManager->PEPManager`, I think. Beyond that I don't see anything else that needs changing.

mcodingbot/plugins/python.py
Lunarmagpie: `SCREAMING_SNAKE_CASE` is more clearly a global so I would stick with it.
Otherwise put it back in the plugin class I guess.

mcodingbot/plugins/python.py
```diff
@@ -29,9 +30,10 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
     return action_row
 
 
-def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
-    url = f"https://peps.python.org/pep-{pep_number:04}/"
-    return f"<{url}>" if hide_embed else url
+@plugin.include
+@tasks.loop(hours=1)
```
qexat, line 34: Honestly PEPs are not updated every hour, it can be a day-checking imo

mcodingbot/plugins/python.py
CircuitSacul: @trag1c what do you think

mcodingbot/plugins/python.py
Lunarmagpie: If we are doing every day a cronjob would definitely be better. That way its not dependent on bot start time and I dont need to import datetime.

qexat:
Run `black` please

mcodingbot/plugins/python.py
qexat: > If we are doing every day a cronjob task would definitely be better. That way its not dependent on bot start time and I dont need to import datetime.

Sure!

mcodingbot/plugins/python.py
```diff
@@ -29,9 +30,10 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
     return action_row
 
 
-def get_pep_link(pep_number: int, *, hide_embed: bool) -> str:
-    url = f"https://peps.python.org/pep-{pep_number:04}/"
-    return f"<{url}>" if hide_embed else url
+@plugin.include
+@tasks.cronjob("@daily")
```
CircuitSacul, line 34: Might be a good idea to set this to a specific time, like when the server is least active? I don't know how much of a load this task takes on the bot, but running it at midnight makes the most sense to me.

If it's too much work to set a specific time dw about it.

qexat: Midnight? UTC?

Lunarmagpie: I dont think theres server activity data. Also time have to be adjusted for the bots timezone so thats already kinda sus.

CircuitSacul: Seems the server is basically inactivate at midnight est

CircuitSacul: *based on messages in #general

CircuitSacul: > Also time have to be adjusted for the bots timezone so thats already kinda sus.

yeah dw about it I guess, it'll be fine as it is

CircuitSacul (approved):

Endercheif (approved):

qexat (approved):

mcodingbot/utils/peps.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import annotations
+
+import dataclasses
```
trag1c, line 3: I'd change this to a from import for consistency

trag1c (changes requested):

mcodingbot/utils/peps.py
qexat: `@dataclasses.dataclass` is ugly + generally you only use `dataclass` from it so I fully agree

CircuitSacul (approved):

trag1c (approved):

qexat:
merging moment

qexat:
I wonder who's gonna merge the fastest

Lunarmagpie:
me

Endercheif:
wait for ci to pass

Endercheif:
:why:

qexat:
> me

:trollface:

PR #53, opened 2022-09-14
handle deletes & edits for pep responses

CircuitSacul:
https://files.circuitsacul.dev/Screen_Recording_2022-09-13_at_8.41.36_PM.mov

### DIFF
--- a/mcodingbot/plugins/peps.py
+++ b/mcodingbot/plugins/peps.py
@@ -15,9 +15,10 @@ from mcodingbot.utils import PEPManager, Plugin
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 DISMISS_BUTTON_ID = "dismiss"
 MAX_AGE_FOR_SEND = timedelta(minutes=1)
+MAX_AGE_FOR_EDIT = timedelta(minutes=5)
 
 recent_pep_responses: TTLCache[int, int] = TTLCache(
-    maxsize=100, ttl=MAX_AGE_FOR_SEND.total_seconds() + 5
+    maxsize=100, ttl=MAX_AGE_FOR_EDIT.total_seconds()
 )
 plugin = Plugin()
 pep_manager = PEPManager()
--- a/mcodingbot/plugins/peps.py
+++ b/mcodingbot/plugins/peps.py
@@ -16,8 +16,7 @@ plugin = Plugin()
 pep_manager = PEPManager()
 MAX_AGE_FOR_SEND = timedelta(minutes=1)
 recent_pep_responses: TTLCache[int, int] = TTLCache(
-    maxsize=100,
-    ttl=MAX_AGE_FOR_SEND.total_seconds() + 5,
+    maxsize=100, ttl=MAX_AGE_FOR_SEND.total_seconds() + 5
 )
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
--- a/mcodingbot/plugins/peps.py
+++ b/mcodingbot/plugins/peps.py
@@ -14,7 +14,11 @@ from mcodingbot.utils import PEPManager, Plugin
 
 plugin = Plugin()
 pep_manager = PEPManager()
-recent_pep_responses: TTLCache[int, int] = TTLCache(maxsize=100, ttl=60 * 5)
+MAX_AGE_FOR_SEND = timedelta(minutes=1)
+recent_pep_responses: TTLCache[int, int] = TTLCache(
+    maxsize=100,
+    ttl=MAX_AGE_FOR_SEND.total_seconds() + 5,
+)
 
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 DISMISS_BUTTON_ID = "dismiss"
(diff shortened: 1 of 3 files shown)


trag1c:
You changed `60 * 5` to `60 + 5`?

CircuitSacul:
> You changed `60 * 5` to `60 + 5`?

Originally I hardcoded it to 5 minutes. Then I decided to just make it 5 seconds longer than the "MAX_AGE_FOR_SEND".

trag1c:
> > You changed `60 * 5` to `60 + 5`?
> 
> Originally I hardcoded it to 5 minutes. Then I decided to just make it 5 seconds longer than the "MAX_AGE_FOR_SEND".

Then just do `timedelta(minutes=1, seconds=5)` or `timedelta(seconds=65)`?

CircuitSacul:
> > > You changed `60 * 5` to `60 + 5`?
> > 
> > 
> > Originally I hardcoded it to 5 minutes. Then I decided to just make it 5 seconds longer than the "MAX_AGE_FOR_SEND".
> 
> Then just do `timedelta(minutes=1, seconds=5)` or `timedelta(seconds=65)`?

No because the TTL is directly based on the MAX_AGE_FOR_SEND constant. It needs to be just a little bit longer than that.

mcodingbot/plugins/peps.py
```diff
@@ -58,31 +67,80 @@ async def callback(self, ctx: crescent.Context) -> None:
...
-        return
+def within_age_cutoff(message_created_at: datetime) -> bool:
+    return datetime.now(timezone.utc) - message_created_at <= MAX_AGE_FOR_SEND
+
 
+def get_peps_embed(content: str) -> hikari.Embed | None:
     pep_refs = [
-        int(ref.groupdict()["pep"])
-        for ref in re.finditer(PEP_REGEX, event.message.content)
+        int(ref.groupdict()["pep"]) for ref in re.finditer(PEP_REGEX, content)
```
trag1c, line 76:
```suggestion
        int(ref.group("pep")) for ref in re.finditer(PEP_REGEX, content)
```

trag1c (changes requested):

trag1c (approved):

mcodingbot/plugins/peps.py
```diff
@@ -1,14 +1,23 @@
...
+from cachetools import TTLCache
 from crescent.ext import tasks
 
 from mcodingbot.config import CONFIG
 from mcodingbot.utils import PEPManager, Plugin
 
 plugin = Plugin()
 pep_manager = PEPManager()
+MAX_AGE_FOR_SEND = timedelta(minutes=1)
+recent_pep_responses: TTLCache[int, int] = TTLCache(
```
Lunarmagpie, line 18: add a whitespace
```suggestion
pep_manager = PEPManager()

MAX_AGE_FOR_SEND = timedelta(minutes=1)
recent_pep_responses: TTLCache[int, int] = TTLCache(
```

Lunarmagpie:
`re`, `crescent`, and `hikari` do not use from imports. This PR should stay consistent with the rest of the file.

Endercheif (approved):

mcodingbot/plugins/peps.py
```diff
@@ -1,14 +1,23 @@
...
+    maxsize=100, ttl=MAX_AGE_FOR_SEND.total_seconds() + 5
```
Lunarmagpie, line 19: The 5's job will be more clear if you do `MAX_AGE_FOR_SEND-5` within the `within_age_cutoff` function

mcodingbot/plugins/peps.py
```diff
@@ -58,31 +67,80 @@ async def callback(self, ctx: crescent.Context) -> None:
         await ctx.respond(embed=pep.embed())
 
 
-@plugin.include
-@crescent.event
-async def on_message(event: hikari.MessageCreateEvent) -> None:
-    if not event.message.content or event.author.is_bot:
-        return
+def within_age_cutoff(message_created_at: datetime) -> bool:
```
Lunarmagpie, line 70:
```suggestion
def is_within_age_cutoff(message_created_at: datetime) -> bool:
```

Lunarmagpie (changes requested):

mcodingbot/plugins/peps.py
CircuitSacul: I have to disagree. The MAX_AGE_FOR_SEND is exactly how old the message is allowed to be for the bot to try and reply to it. MAX_AGE_FOR_SEND-5 is 5 seconds less than that.

It would be like doing
```py
LOOP_DELAY

while True:
    sleep(LOOP_DELAY-5)
```

mcodingbot/plugins/peps.py
```diff
@@ -1,17 +1,26 @@
...
 
 from mcodingbot.config import CONFIG
 from mcodingbot.utils import PEPManager, Plugin
 
-plugin = Plugin()
-pep_manager = PEPManager()
-
 PEP_REGEX = re.compile(r"pep[\s-]*(?P<pep>\d{1,4}\b)", re.IGNORECASE)
 DISMISS_BUTTON_ID = "dismiss"
+MAX_AGE_FOR_SEND = timedelta(minutes=1)
```
Lunarmagpie, line 17: I think 2 minutes is a better choice here. I doubt that cache will get that full so memory shouldn't be a problem + editing a message after 1 minute is not that crazy if the channel isn't really moving.
```suggestion
MAX_AGE_FOR_SEND = timedelta(minutes=2)
```

mcodingbot/plugins/peps.py
CircuitSacul: feels wordy and unnecessary imo

mcodingbot/plugins/peps.py
Lunarmagpie: I see the `within_cutoff` function is asking "is this a safe time to send the message" so it should do adjustments to ensure that there

CircuitSacul: Yeah, doing 5 seconds less doesn't make sense.

Lunarmagpie (approved):

PR #57, opened 2022-09-14
Make class commands into functions

Lunarmagpie:
removes an unnecessary parameter and indentation level. I think the code is also clearer when it doesn't have a class that doesn't really do any class things.

And its prettier tbh.

### DIFF
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -2,20 +2,20 @@ import crescent
 import hikari
 
 from mcodingbot.config import CONFIG
-from mcodingbot.utils import Plugin
+from mcodingbot.utils import Context, Plugin
 
 plugin = Plugin()
 
 
 @plugin.include
 @crescent.command(name="ping", description="Pong!")
-async def ping(ctx: crescent.Context) -> None:
+async def ping(ctx: Context) -> None:
     await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
 
 
 @plugin.include
 @crescent.command(name="links", description="Useful links")
-async def links(ctx: crescent.Context) -> None:
+async def links(ctx: Context) -> None:
     embed = hikari.Embed(
         title="Links",
         color=CONFIG.theme,
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -9,22 +9,20 @@ plugin = Plugin()
 
 @plugin.include
 @crescent.command(name="ping", description="Pong!")
-class PingCommand:
-    async def callback(self, ctx: crescent.Context) -> None:
-        await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
+async def ping(ctx: crescent.Context) -> None:
+    await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
 
 
 @plugin.include
 @crescent.command(name="links", description="Useful links")
-class LinksCommand:
-    async def callback(self, ctx: crescent.Context) -> None:
-        embed = hikari.Embed(
-            title="Links",
-            color=CONFIG.theme,
-            description=(
-                f"**[mCoding Youtube]({CONFIG.mcoding_youtube})**\n"
-                f"**[mCoding repo]({CONFIG.mcoding_repo})**\n"
-                f"**[mCodingBot repo]({CONFIG.mcodingbot_repo})**"
-            ),
-        )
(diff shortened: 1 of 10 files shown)


CircuitSacul:
My personal opinion is to only use class-commands for consistency. But if @trag1c agrees with you, then I'm fine with this change.

trag1c:
Is there any benefit to using classes?

Lunarmagpie:
> Is there any benefit to using classes?

Readability for command options. So none in this case.

trag1c:
> > Is there any benefit to using classes?
> 
> Readability for command options. So none in this case.

Go for functions then

Lunarmagpie:
class commands are slower than function commands as well :trollface:

CircuitSacul:
> Is there any benefit to using classes?
Consistency, since all commands currently use classes. This PR makes it so that only commands with options have classes.

Endercheif (approved):

Lunarmagpie:
Bump

trag1c (approved):

mcodingbot/plugins/info.py
```diff
@@ -9,22 +9,20 @@
 
 @plugin.include
 @crescent.command(name="ping", description="Pong!")
-class PingCommand:
-    async def callback(self, ctx: Context) -> None:
-        await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
+async def ping(ctx: Context) -> None:
+    await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
```
qexat, line 13:
```suggestion
    await ctx.respond(f"Pong! {int(ctx.app.heartbeat_latency*1000)} ms.")
```
`round()` has the weird behavior to round to the nearest even, you might not like that
Also, it's personal tastes, but I like when there are spaces around the operator `*` and it follows PEP 8 since it is technically the operator with the lowest priority (I mean, it's alone)

mcodingbot/plugins/info.py
```diff
@@ -9,22 +9,20 @@
...
-            ),
-        )
-        await ctx.respond(embed=embed)
+async def links(ctx: Context) -> None:
+    embed = hikari.Embed(
+        title="Links",
+        color=CONFIG.theme,
+        description=(
+            f"**[mCoding Youtube]({CONFIG.mcoding_youtube})**\n"
+            f"**[mCoding repo]({CONFIG.mcoding_repo})**\n"
```
qexat, line 24: Just a thought, wouldn't be better to write `repository` in full?
```suggestion
            f"**[mCoding repository]({CONFIG.mcoding_repo})**\n"
```

mcodingbot/plugins/info.py
```diff
@@ -9,22 +9,20 @@
...
-                f"**[mCodingBot repo]({CONFIG.mcodingbot_repo})**"
-            ),
-        )
-        await ctx.respond(embed=embed)
+async def links(ctx: Context) -> None:
+    embed = hikari.Embed(
+        title="Links",
+        color=CONFIG.theme,
+        description=(
+            f"**[mCoding Youtube]({CONFIG.mcoding_youtube})**\n"
```
qexat, line 23: Also, I suggest changing `Youtube` to `YouTube`
```suggestion
            f"**[mCoding YouTube]({CONFIG.mcoding_youtube})**\n"
```

qexat (changes requested):

mcodingbot/plugins/info.py
qexat: or even `YouTube channel`

Lunarmagpie: this is outside the scope of this pr

mcodingbot/plugins/info.py
Lunarmagpie: Outside scope

mcodingbot/plugins/info.py
Lunarmagpie: Outside scope

mcodingbot/plugins/info.py
qexat: That's a very small change though

Lunarmagpie: Then pr it after. This PR modifies how three functions are written. It will not do anything else.

Lunarmagpie: Adding more to this PR will a) make reviews harder and b) add other issues that need to be resolved making it take longer to merge

qexat: alright alright

Endercheif (approved):

PR #60, opened 2022-09-14
Make things consistent

### LINKED ISSUE #50: Move custom Plugin, use it consistently, custom context
I just realized that some plugins are using crescent.Plugin, while others are using the one located in utils/. That should be fixed, and while we're at it, I think it should be moved.

These are the tasks I have in mind, idk if we want to do all of them or not.
 - [ ] Move utils.plugin:Plugin to plugins._plugin:Plugin
 - [ ] Always use the subclassed version of the plugin
 - [ ] Add a custom context that also casts the type of bot

Lunarmagpie:
close #50


Lunarmagpie:
me making a merge conflict with my other pr :trollface:

CircuitSacul (approved):

CircuitSacul (changes requested):
Actually I don't like the filename change. I think it makes more sense to leave it as `plugin.py`

what I mean was we should move it to a `_plugin.py` file inside of `plugins/`, but inside utils is fine as well.

Lunarmagpie:
I dont agree with moving it to the plugins folder because I think the project will be easier to read if nothing is imported from there.
I can change the filename back tomorrow

CircuitSacul:
Alright for now just leave it where it is and change its name back to `plugin`. We can move files around later if we want to.

CircuitSacul (approved):

trag1c (approved):

Issue #55, opened 2022-09-14
Add search query to /pep command


CircuitSacul:
The /pep command can have two optional arguments:
 - number
 - query

You must specify exactly one option. Number just returns that pep, whereas query will use fuzzy searching to search the peps by their titles. This is also the easiest way to introduce fuzzy searching to the codebase.

Endercheif:
I can do this

Endercheif:
Already experimenting with it.

isFakeAccount:
Can somone also create a command that lets you fuzzy search mcoding videos and return link.

trag1c:
> Can somone also create a command that lets you fuzzy search mcoding videos and return link.

Maybe read through the issues first :/ (#54)

Endercheif:
I'm going to wait until #57 is merged before this.

Endercheif:
nvm i'll do it now

PR #62, opened 2022-09-15
`/pep` autocomplete

Endercheif:


Endercheif:
`poetry.lock` :sob: how to fix

Endercheif:
Implements #55

mcodingbot/plugins/peps.py
```diff
@@ -30,6 +30,23 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
     return action_row
 
 
+async def autocomplete_pep(
+    ctx: crescent.AutocompleteContext,
+    options: hikari.AutocompleteInteractionOption,
```
Lunarmagpie, line 45: should be `option` (no s)

mcodingbot/plugins/peps.py
```diff
@@ -41,17 +58,14 @@ async def update_peps() -> None:
     name="pep", description="Find a Python Enhancement Proposal."
 )
 class PEPCommand:
-    pep_number = crescent.option(
-        int, "The PEP number.", name="pep-number", min_value=0, max_value=9999
-    )
-    show_embed = crescent.option(
-        bool, "Whether to show the embed.", name="show-embed", default=True
+    pep = crescent.option(
+        int, "PEP number/title", autocomplete=autocomplete_pep
```
Lunarmagpie, line 72:
```suggestion
        int, "The PEP number or title.", autocomplete=autocomplete_pep
```

mcodingbot/utils/peps.py
```diff
@@ -2,12 +2,13 @@
...
 from dataclasses import dataclass
 from logging import getLogger
-from typing import TYPE_CHECKING, Sequence
+from typing import TYPE_CHECKING, Sequence, Generator
 
 import aiohttp
 import hikari
 
 from mcodingbot.config import CONFIG
+from mcodingbot.utils.search import find
```
Lunarmagpie, line 11:
```suggestion
from mcodingbot.utils import find
```
Its re-exported

mcodingbot/utils/peps.py
```diff
@@ -39,10 +41,19 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
+                self._pep_map = {
+                    k: f"{v.title} ({v.number})" for k, v in self._peps.items()
+                }
 
     def get(self, pep_number: int) -> PEPInfo | None:
         return self._peps.get(pep_number)
 
+    def search(
+        self, query: str, limit: int = 20
+    ) -> Generator[PEPInfo, None, None]:
```
Lunarmagpie, line 53: This returns `typing.Iterable[PEPInfo]`

mcodingbot/utils/peps.py
```diff
@@ -39,10 +41,19 @@ async def fetch_pep_info(self, bot: Bot) -> None:
                     )
                     for pep_id, pep in (await resp.json()).items()
                 }
+                self._pep_map = {
+                    k: f"{v.title} ({v.number})" for k, v in self._peps.items()
+                }
```
Lunarmagpie, line 46: Gonna need to think about this. I feel like this can be done without a second dict.

Lunarmagpie (changes requested):

mcodingbot/utils/peps.py
CircuitSacul: No it returns a generator. Generators are also Iterables, but it's best to be specific with returns types and broad with input types.

mcodingbot/utils/peps.py
CircuitSacul: The search function accepts a very specific format (either Collection[str] or Mapping[value, str]). Passing the original dict wouldn't work, and passing just the title would make it impossible to get the rest of the pep info from the response.

The dict either needs to be generated every time this function is called, or stored separately.

I could be missing something ofc.

mcodingbot/utils/peps.py
Endercheif: Keep it or change it?

mcodingbot/utils/search.py
```diff
@@ -0,0 +1,24 @@
...
+from collections.abc import Mapping
+from typing import TypeVar
+
+from rapidfuzz import fuzz, process
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+def find(
```
CircuitSacul, line 12: I feel like we need a more descriptive name maybe. Perhaps `fuzzysearch`?

mcodingbot/utils/peps.py
CircuitSacul: I say keep it, unless @Lunarmagpie has a good reason to change it

mcodingbot/utils/peps.py
Endercheif: I spent a while on thinking of a better way to do this. Unfortunately this is the best way other than generating it every time it is called.

mcodingbot/utils/peps.py
```diff
@@ -39,10 +41,19 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
 
     def get(self, pep_number: int) -> PEPInfo | None:
         return self._peps.get(pep_number)
 
+    def search(
+        self, query: str, limit: int = 20
+    ) -> Generator[PEPInfo, None, None]:
+        res = find(query, self._pep_map, limit=limit)
+        return (self.get(pep[2]) for pep in res)
```
CircuitSacul, line 55: Don't forget to fix this part (this is why mypy is failing). One way to fix it is like this:

```py
return (
    pep_info
    for pep in res
    if (pep_info := self.get(pep[2])) is not None
)

File: mcodingbot/utils/peps.py (hunk 4 above)
Comment (Lunarmagpie) [reply mcodingbot/utils/peps.py:53]: Generator inherits from Iterable (check typeshed)
we dont need any of the added things. i do not believe they apply when its Generator[T, None, None]

Comment (CircuitSacul) [reply mcodingbot/utils/peps.py]: If generator inherits from Iterable then this should be a Generator, as that's what's being returned.

You wouldn't do this
```py
def function() -> object:
    return "hello"
```

You would use `str`, which is a subclass of `object`, as that is more specific. Your return types should be as specific as possible, and argument types should be as broad as possible.

Lunarmagpie: nobody returns a string to use it like an object...
i will correct myself `Iterator` is more accurate. `Generator[T, None, None`] does not give us a single thing that does shit over Iterator.  I'm sure throw, close, gi_code, gi_frame, gi_yieldfrom, and gi_running are real useful. Your argument assumes they are as useful as basic string methods.

CircuitSacul: Generators behave completely differently from all other Iterables, as Generators are consumable.

CircuitSacul: https://discord.com/channels/826875707418214451/851854179534372884/1019844894912675892

Endercheif: @trag1c thoughts ~~save me~~

Lunarmagpie: > 

So use iterator. The docs recommend it as an alternative as well.

mcodingbot/utils/search.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import annotations
+
+from collections.abc import Mapping
```
Lunarmagpie, line 3: Use typing mapping for consistentcy

Lunarmagpie (changes requested):

mcodingbot/utils/peps.py
trag1c: > @trag1c thoughts? ~save me~

Use `Iterator[PEPInfo]`

mcodingbot/utils/peps.py
```diff
@@ -39,10 +41,23 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
 
+    def search(
+        self, query: str, limit: int = 20
+    ) -> Generator[PEPInfo, None, None]:
+        res = fuzzy_search(query, self._pep_map, limit=limit)
+        return (
+            pep_info
+            for pep in res
+            if (pep_info := self.get(pep[2])) is not None
+        )
```
trag1c, line 59: Why not just use `yield` in a normal loop? Would be more readable imo (mainly because of the walrus)

Lunarmagpie: +1

Lunarmagpie (changes requested):

mcodingbot/utils/peps.py
Lunarmagpie: +1 for tragic

CircuitSacul: A yield seems like a good idea.

```py
def search(...) -> Generator[PEPInfo, None, None]:
    res = ...
    for pep in res:
        if (pep_info := self.get(pep[2])):
            yield pep
```

Lunarmagpie (approved):

CircuitSacul (approved):

mcodingbot/utils/search.py
```diff
@@ -0,0 +1,23 @@
...
+
+from typing import Mapping, TypeVar
+
+from rapidfuzz import fuzz, process
+
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+def fuzzy_search(
```
Lunarmagpie, line 11: Removes default values being assigned in different places. They are already different so it's an issue.
```python
def fuzzy_search(
    query: str,
    choices: Mapping[K, V],
    *,
    score_cutoff: int | float | None,
    limit: int,
) -> list[tuple[V, float, K]]:
```

Endercheif: can you make this into a suggestion pls?

Lunarmagpie: No. I have locked myself out of the room with my computer so I'm a bit fucked rn.

mcodingbot/utils/search.py
```diff
@@ -0,0 +1,23 @@
...
+    query: str,
+    choices: Mapping[K, V],
+    score_cutoff: int | float | None = None,
+    limit: int = 10,
+) -> list[tuple[V, float, K]]:
```
trag1c, line 16:
```suggestion
def fuzzy_search(
    query: str,
    choices: Mapping[K, V],
    *,
    score_cutoff: int | float | None,
    limit: int,
) -> list[tuple[V, float, K]]:
```

mcodingbot/utils/peps.py
```diff
@@ -39,10 +41,19 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
                     for pep_id, pep in (await resp.json()).items()
                 }
+                self._pep_map = {
+                    k: f"{v.title} ({v.number})" for k, v in self._peps.items()
+                }
 
     def get(self, pep_number: int) -> PEPInfo | None:
         return self._peps.get(pep_number)
 
+    def search(self, query: str, limit: int = 20) -> Iterator[PEPInfo]:
```
Lunarmagpie, line 51:
```suggestion
    def search(self, query: str, *, limit: int) -> Iterator[PEPInfo]:
```

mcodingbot/plugins/peps.py
```diff
@@ -40,6 +40,23 @@ def get_dismiss_button(id: hikari.Snowflake) -> hikari.api.ActionRowBuilder:
...
 
 
+async def autocomplete_pep(
+    ctx: crescent.AutocompleteContext,
+    option: hikari.AutocompleteInteractionOption,
+) -> list[hikari.CommandChoice]:
+    if not option.value:
+        return []
+
+    results = pep_manager.search(option.value)
```
Lunarmagpie, line 50:
```suggestion
    results = pep_manager.search(option.value, limit=20)
```

Lunarmagpie (changes requested):
I don't think it makes more sense to abstract the limit out of the autocomplete function.

mcodingbot/utils/search.py
```diff
@@ -0,0 +1,24 @@
...
+K = TypeVar("K")
+V = TypeVar("V")
+
+
+def fuzzy_search(
+    query: str,
+    choices: Mapping[K, V],
+    *,
+    score_cutoff: int | float | None,
+    limit: int,
```
CircuitSacul, line 16: kw-only arguments should have a default value imo.

Just make both of these None, I think?

CircuitSacul: wait lunar just asked you to change that lol. hm

CircuitSacul: okay yeah if you can make this arguments None that would be great.

Endercheif: @CircuitSacul leave a suggestion then i will change it

Lunarmagpie (approved):

trag1c (approved):

Endercheif:
🚀🚀🚀

PR #64, opened 2022-09-15
fix: use custom helper function instead of `round`

qexat:
# Fix: use custom helper function instead of `round`

## Incipit

In Python 3.0, **`round()` built-in behavior changed** to follow industry standards by using [Banker's rounding algorithm](https://en.wikipedia.org/wiki/Rounding#Round_half_to_even) (which eliminates some mathematical bias), as stipulated by [this changelog](https://docs.python.org/3/whatsnew/3.0.html#:~:text=The%20round()%20function%20rounding%20strategy%20and%20return%20type%20have%20changed.).

## Consequences

This means that `round()` **rounds to the nearest even**, like the "Bankers" column in this table:
image

Even though this is a great thing it happened, as it distributes proportionally the roundings and as [it also follows IEEE 754 as well](https://en.wikipedia.org/wiki/IEEE_754#Roundings_to_nearest), which is in favor of a global programming consistency, this rounding method is neither natural nor **expected in the context** it is currently applied in the code.

## Replacement candidate

`int()` built-in type could be a great candidate to replace `round()` ; however, it has one issue: **it always rounds down**, which is not natural rounding.

## Decision

Thus, my decision was to **write my own helper function**, which has to be simple and efficient.

Mimicking Python's own style, here's what I wrote:

```py
def natural_round(x: float) -> int:
    n = int(x)
    return n + (x - n >= .5)
```

Edited according to recent commits.

## Performance

I did not deeply benchmarked it, but on average, it's about **3 times slower than `int`** with large numbers (from `-1_000_000` to `1_000_000`, 1 million times).

Here's the [source code](https://pastebin.com/EmTKLVPU) of the benchmark script I've used.

On my machine, `int()` takes about `0.06` seconds whereas `nat_round()` takes about `0.19` seconds, which makes sense as this latter calls the first.

## Conclusion

I have therefore chosen to propose this PR, in order to fix this potentially unintended behavior.

Indeed:

- **`round()` is not really required** anyway, as we don't need/want specifically to round to a specific number of digits but **to the nearest integer**.
- `int()` does not fix the issue.
- `nat_round()` makes the rounding natural, as we would expect from data like latency.

### DIFF
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -7,11 +7,16 @@ from mcodingbot.utils import Context, Plugin
 plugin = Plugin()
 
 
+def nat_round(__x: int | float) -> int:
+    n = int(__x)
+    return n + (__x - n >= .5)
+
+
 @plugin.include
 @crescent.command(name="ping", description="Pong!")
 class PingCommand:
     async def callback(self, ctx: Context) -> None:
-        await ctx.respond(f"Pong! {int(ctx.app.heartbeat_latency * 1000)} ms.")
+        await ctx.respond(f"Pong! {nat_round(ctx.app.heartbeat_latency * 1000)} ms.")
 
 
 @plugin.include
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -11,7 +11,7 @@ plugin = Plugin()
 @crescent.command(name="ping", description="Pong!")
 class PingCommand:
     async def callback(self, ctx: Context) -> None:
-        await ctx.respond(f"Pong! {round(ctx.app.heartbeat_latency*1000)} ms.")
+        await ctx.respond(f"Pong! {int(ctx.app.heartbeat_latency * 1000)} ms.")
 
 
 @plugin.include
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -7,15 +7,22 @@ from mcodingbot.utils import Context, Plugin
 plugin = Plugin()
 
 
-def nat_round(__x: float) -> int:
-    n = int(__x)
-    return n + (__x - n >= 0.5)
+def natural_round(x: float) -> int:
+    """
+    Rounds in the same way as Python 2's `round()`.
+    If the real part of x is >= .5, rounds up ; else, rounds down.
+    """
+
+    n = int(x)
+    return n + (x - n >= 0.5)
 
 
 @plugin.include
 @crescent.command(name="ping", description="Pong!")
 async def ping(ctx: Context) -> None:
-    await ctx.respond(f"Pong! {nat_round(ctx.app.heartbeat_latency*1000)} ms.")
+    await ctx.respond(
+        f"Pong! {natural_round(ctx.app.heartbeat_latency*1000)} ms."
+    )
 
 
 @plugin.include
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -7,7 +7,7 @@ from mcodingbot.utils import Context, Plugin
 plugin = Plugin()
 
 
-def nat_round(__x: int | float) -> int:
+def nat_round(__x: float) -> int:
     n = int(__x)
     return n + (__x - n >= 0.5)
 
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -9,7 +9,7 @@ plugin = Plugin()
 
 def nat_round(__x: int | float) -> int:
     n = int(__x)
-    return n + (__x - n >= .5)
+    return n + (__x - n >= 0.5)
 
 
 @plugin.include


mcodingbot/plugins/info.py
```diff
@@ -7,10 +7,15 @@
 plugin = Plugin()
 
 
+def nat_round(__x: float) -> int:
```
Lunarmagpie, line 10:
```suggestion
def natural_round(__x: float) -> int:
```
could use a docstring on why its round. Something along the lines of. `"""rounds .5 up instead of down like round()"""`

Lunarmagpie (changes requested):

mcodingbot/plugins/info.py
```diff
@@ -7,10 +7,15 @@
...
+    n = int(__x)
+    return n + (__x - n >= 0.5)
```
trag1c, line 12: Why `__x`?

qexat: Python style idk

qexat: Yeah i'll just put `x`

qexat:
Updated the title as it wasn't conforming to actual PR's content

Endercheif:
Is this PR even needed? We do not need for the latency to be precise. When people use `/ping` they are either doing it just because they want to or to see an approximate. Values of +-1 are not a issue like it is made to be. 
It would make sense for this PR if we were monitoring the latency over a period of time but alas we are not.

qexat:
> Values of +-1 are not a issue like it is made to be.

This is true when the ping is above 100ms, but I wouldn't really agree when it's lower.
My PR is also to point out a specific behavior, as `round()` could have been used in the future in other places.

Lunarmagpie:
I agree with endercheif. The cost of this is kinda high for what we get out of it.

trag1c:
> I agree with endercheif. The cost of this is kinda high for what we get out of it.

How is the cost high?

Lunarmagpie:
Readability cost of an unexpected function /shrug
i dont care about performance

qexat:
> Readability cost of an unexpected function /shrug i dont care about performance

How the function makes the code less readable? :raised_eyebrow:

Lunarmagpie:
> How the function makes the code less readable? raised_eyebrow

Reader has to check why regular round isn't being used

qexat:
> > How the function makes the code less readable? raised_eyebrow
> 
> Reader has to check why regular round isn't being used

Well this could just be added to helper's docstring :woman_shrugging:

Lunarmagpie (approved):
i can't live knowing that the answer could be wrong. is merging the right choice? idk but its gonna annoy me

qexat:
> i can't live knowing that the answer could be wrong. is merging the right choice? idk but its gonna annoy me

Let's keep it a draft for now. If you will to merge I'll just undraft it

mcodingbot/plugins/info.py
```diff
@@ -7,10 +7,22 @@
...
 
 
+def natural_round(x: float) -> int:
+    """
+    Rounds in the same way as Python 2's `round()`.
+    If the real part of x is >= .5, rounds up ; else, rounds down.
+    """
+
+    n = int(x)
+    return n + (x - n >= 0.5)
```
Sigmanificient, line 17: What about using a simple ternary?
```suggestion
    return x + 1 if x % 1 == .5 else x
```

```
nat_round_mod_ternary: executed 1000000 times in 0.14679307599999447 seconds.
```
# 

<br>

Here is an another way to write it, without the mod operator:
```suggestion
    return (y := int(x)) + (x - y >= .5)
```
```
nat_round_walrus_minus: executed 1000000 times in 0.14819169599991255 seconds.
```


other results (for reference):
```
round: executed 1000000 times in 0.07041415899993808 seconds.
int: executed 1000000 times in 0.04023135199986427 seconds.
nat_round: executed 1000000 times in 0.152997527000025 seconds.
```

Sigmanificient: I love over-engineer features :D

qexat:
```suggestion
    return round(2 * x + .5) >> 1
```
is also a possibility. What do you think @Sigmanificient?

qexat: > How is that last one working? xD

`round()` rounds 4.5 to 4 but 5.5 to 6.
If you multiply `x` by 2, you get rid of this even preference, and `+ .5` is to force rounding (it will disappear anyway since it rounds)
And the `>> 1` is just a fast `// 2` on integers

So for example:
`x = 4.5`
`round(9 + .5)` <=> `round(9.5)` == 10
10 >> 1 <=> 10 // 2 == 5, 4.5 rounded is 5

`x = 5.5`
`round(11 + .5)` <=> `round(11.5)` == 12
12 >> 1 == 6, 5.5 rounded is 6

Problem fixed :sunglasses: :+1:

Sigmanificient: well i guess you won that one x)

qexat: or you could even use `int(x+.5)`

Sigmanificient: Sad that `&` and `|` dont work with float, could have been fun to mess with that

trag1c: > or you could even use `int(x+.5)`

doesn't work for negative numbers btw

qexat: > > or you could even use `int(x+.5)`
> 
> doesn't work for negative numbers btw

nor my solution xd

qexat:
```suggestion
    if not n:
        return n
    return round(2 * n + n // abs(n) * .5) >> 1
```

Sigmanificient: that last once seems slower with use of `//`? What does thez benchmark says about it?

qexat: > that last one seems slower with use of `//`? What does the benchmark says about it?

way slower actually

Sigmanificient: Well my evil one is quite slow too
```
nat_round_evil: executed 1000000 times in 0.2027178480002476 seconds.
```

Lunarmagpie: How about we just don't change it?

CircuitSacul:
Why close?

qexat:
> Why close?

Ask @Lunarmagpie

Lunarmagpie:
I meant that comment thread to change the impl of `natural_round` was about a problem that didn't need changing. Its fine how it is.

PR #65, opened 2022-09-15
Better numerical PEP search, fix PEP 499 too long

Lunarmagpie:
If the query is a number return numbers that match the given number first. The rest remaining on limit is filled in with fuzzy sort.


Lunarmagpie:
Rip there's a logic issue with this

Lunarmagpie:
Kinda broke one of my rules and fixed a bug that prevented autocomplete from working for pep 499

Lunarmagpie:
bump

mcodingbot/utils/peps.py
```diff
@@ -48,13 +49,37 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
+
         res = fuzzy_search(query, self._pep_map, limit=limit)
         for pep in res:
             if pep_info := self.get(pep[2]):
+                if pep_info in items:
+                    continue
+                yielded += 1
                 yield pep_info
+                if limit and yielded >= limit:
+                    return
```
CircuitSacul, line 82: This could theoretically return more than the limit, if there are > entries that match the input numerically.

mcodingbot/utils/peps.py
```diff
@@ -71,5 +96,16 @@ def embed(self) -> hikari.Embed:
...
+        pep_digits = len(str(self.number))
+
+        max_name_length = 100 - pep_digits - 3
+
+        name = self.title
+        if len(name) > max_name_length:
+            name = f"{self.title[:max_name_length - 3]}..."
+
+        return f"{name} ({self.number})"
+
```
CircuitSacul, line 109: I see `-3` twice... Also 100 looks like a magic number.

At least some comments would be nice.

CircuitSacul (changes requested):

mcodingbot/utils/peps.py
Lunarmagpie: True it is a magic number, ill add some comments.

mcodingbot/utils/peps.py
Lunarmagpie: Not possible. `_get_matches_digits` returns `limit` entries max.

Lunarmagpie:
made the changes

mcodingbot/utils/peps.py
CircuitSacul, line 82: This can still return 1 more than the limit. The `yielded` check is done after the first yield.

So, return `limit` from `_get_matches_digits`, and then potentially return one more from the main loop.

CircuitSacul (changes requested):

mcodingbot/utils/peps.py
Lunarmagpie: fixed

mcodingbot/utils/peps.py
```diff
@@ -48,12 +49,36 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
+            items, yielded = self._get_matches_digits(query, limit)
+            yield from items
+
         res = fuzzy_search(query, self._pep_map, limit=limit)
         for pep in res:
             if pep_info := self.get(pep[2]):
+                if pep_info in items:
+                    continue
+                if limit and yielded >= limit:
+                    return
```
CircuitSacul, line 80: Put the limit check before the `continue`? Potentially saving a few loops.

CircuitSacul (changes requested):

CircuitSacul:
Title suggestion: "Better numerical PEP search, fix PEP 499 too long"

mcodingbot/utils/peps.py
```diff
@@ -48,12 +49,36 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
+
         res = fuzzy_search(query, self._pep_map, limit=limit)
         for pep in res:
             if pep_info := self.get(pep[2]):
+                if limit and yielded >= limit:
+                    return
+                if pep_info in items:
+                    continue
+                yielded += 1
                 yield pep_info
```
CircuitSacul, line 82: Just had a thought. Would it be better to add a check right after line 72:
```py
if yielded >= limit:
    return
```

That way the fuzzy search is never reached? Then the limit check and the `continue` can be flipped back.

Lunarmagpie: Nah, I would rather not put the same check in twice.

Lunarmagpie: ended up doing this

Lunarmagpie:
Think im happy with the logic now. Bumped the amount autocompleted to 25 (max) since it looks fine on discord.

mcodingbot/utils/peps.py
```diff
@@ -48,13 +49,43 @@ async def fetch_pep_info(self, bot: Bot) -> None:
...
-                yield pep_info
+            pep_info = self.get(pep[2])
+            if not pep_info:
+                return
+            if limit and yielded >= limit:
+                return
+            if pep_info in items:
+                continue
+            yielded += 1
+            yield pep_info
```
CircuitSacul, line 88: I don't feel like this should fail if somehow one of the peps no longer exists in the dict. honestly this could raise an error, but I think a continue makes more sense (and maybe a _LOG.warn)

```py
        for pep in res:
            pep_info = self.get(pep[2])
            if not pep_info or pep_info in items:
                continue

            yield pep_info
            yielded += 1

            if limit and yielded >= limit:
                return
```

CircuitSacul (changes requested):

mcodingbot/utils/peps.py
Lunarmagpie: done

mcodingbot/utils/peps.py
```diff
@@ -48,13 +49,45 @@ async def fetch_pep_info(self, bot: Bot) -> None:
     def get(self, pep_number: int) -> PEPInfo | None:
         return self._peps.get(pep_number)
 
+    def _get_matches_digits(
+        self, query: str, limit: int | None
+    ) -> tuple[Iterable[PEPInfo], int]:
```
CircuitSacul, line 54: since this returns a list, you should just type it as `list[PEPInfo]`.

CircuitSacul (changes requested):

Lunarmagpie:
This is the longest review for a short PR lol. The change is made.

CircuitSacul (approved):

mcodingbot/utils/peps.py
```diff
@@ -71,5 +104,21 @@ def embed(self) -> hikari.Embed:
             color=CONFIG.theme,
         ).set_author(name=self.authors)
 
+    @property
+    def truncated_title(self) -> str:
+        pep_digits = len(str(self.number))
+
+        # 3 is the length two parenthesis and the space used to seperate the
```
trag1c, line 111:
```suggestion
        # 3 is the length of two parentheses and the space used to separate the
```

mcodingbot/utils/peps.py
```diff
@@ -71,5 +104,21 @@ def embed(self) -> hikari.Embed:
...
+        # pep number from the pep title.
+        max_name_length = 100 - pep_digits - 3
+
+        name = self.title
+        if len(name) > max_name_length:
+            # An extra 3 chars need to be removed to make space for the
```
trag1c, line 117:
```suggestion
            # An extra 3 chars needed to be removed to make space for the
```

trag1c (approved):

PR #69, opened 2022-09-25
Gitignore lockfile

CircuitSacul:
Doing this solves the problem of merge conflicts that can't be resolved manually.


Lunarmagpie (approved):

trag1c (approved):

qexat:
https://python-poetry.org/docs/basic-usage/
explicitely says that `poetry.lock` should be committed.

If you have merge conflicts, try to commit the file separately.
If it doesn't fix the issue, checkout to main branch, then do `poetry lock --no-updates`.

CircuitSacul:
> [Poetry docs](https://python-poetry.org/docs/basic-usage/) explicitely say that `poetry.lock` should be committed.
> 
> If you have merge conflicts, try to commit the file separately. If it doesn't fix the issue, checkout to main branch, then do `poetry lock --no-update`.
> 
> If you're not satisfied with this solution, then commit `poetry.lock` only when it gets updated.
> 
> EDIT: fixed the typo in the lock command flag.

Personally I think that they're just wrong. It causes pain, and the "benefit" is that development is easier because we have the same versions... except that the pyproject.toml file is already very restrictive.

Lunarmagpie:
Although having a lockfille is correct, I think removing it reduces friction for new contributors.

PR #70, opened 2022-09-25
Basic highlight feature

Lunarmagpie:
- [X] Add/remove highlights
- [X] Cache highlights
- [X] dm user
- [x] Pretty display for `highlights list` and highlights dm (assigned to @CircuitSacul)
- [x] Consider different logic for `User.delete_word`
- [x] Investigate typing with `fetchmany`
- [ ] ~~Support regex~~ moving to different PR
- [x] Decide to limit highlights or not (limited to 25)


mcodingbot/database/models/user.py
```diff
@@ -18,3 +24,65 @@ async def get_or_create(user_id: int) -> User:
...
+    @staticmethod
+    async def delete_word(word: str, user_id: int) -> bool:
+        """
+        Return `True` if the word was successfully removed from the user.
+        Return `False` otherwise.
+        """
+        if not await User.exists(user_id=user_id):
+            return False
+
+        user = await User.fetch(user_id=user_id)
```
CircuitSacul, line 43: User.exists returns Optional[User], not a bool.

mcodingbot/database/models/user.py
```diff
@@ -18,3 +24,65 @@ async def get_or_create(user_id: int) -> User:
...
+        words = await user.words.fetchmany()
+
+        for word_ in words:
+            if word_.word == word:
+                break
+        else:
+            return False
```
CircuitSacul, line 50: This is pointless, just use the fetch to check.

mcodingbot/database/models/user.py
```diff
@@ -18,3 +24,65 @@ async def get_or_create(user_id: int) -> User:
...
+
+        try:
+            word_model = await Word.fetch(word=word)
+        except ModelNotFound:
+            return False
```
CircuitSacul, line 55:
```py
if not (word_model := await Word.exists(word=word)):
    return False
```

mcodingbot/database/models/user.py
```diff
@@ -18,3 +24,65 @@ async def get_or_create(user_id: int) -> User:
...
+
+        await user.words.remove(word_model)
+        return True
+
+
+@final
```
CircuitSacul, line 61: I don't think we need this.

mcodingbot/database/models/user.py
```diff
@@ -18,3 +24,65 @@ async def get_or_create(user_id: int) -> User:
...
+
+
+class UserWord(Model):
+    word_id = types.Serial().field()
+    user_id = types.Numeric().field().with_converter(NumericConverter)
+
+    word_id_fk = ForeignKey(word_id, Word.id)
+    user_id_fk = ForeignKey(user_id, User.user_id)
+
+    primary_key = (word_id, user_id)
```
CircuitSacul, line 88: Each model should really be in it's own file, I think.

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,112 @@
...
+from mcodingbot.database.models.user import Word, User
+
+plugin = Plugin()
+highlights_group = crescent.Group("highlights")
+highlights_cache: dict[str, list[hikari.Snowflake]] = defaultdict(list)
+
+
+def _cache_highlight(word: str, *user_ids: hikari.Snowflake) -> None:
+    for user_id in user_ids:
+        highlights_cache[word].append(user_id)
```
CircuitSacul, line 18: .extend()

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,112 @@
...
+        # Deletes empty arrays from the cache
+        if not highlights_cache[word]:
+            del highlights_cache[word]
+
+
+@plugin.include
+@highlights_group.child
+@crescent.command(name="create")
+class CreateHighlight:
+    word = crescent.option(str)
```
CircuitSacul, line 33: Missing descriptions

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,112 @@
...
+@plugin.include
+@highlights_group.child
+@crescent.command(name="delete")
+class DeleteHighlight:
+    word = crescent.option(str)
+
+    async def callback(self, ctx: Context) -> None:
+        if await User.delete_word(self.word, ctx.user.id):
+            await ctx.respond(f'Removed "{self.word}" from your highlights.')
+            _uncache_highlight(self.word, ctx.user.id)
```
CircuitSacul, line 54: uncache before responding

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,112 @@
...
+
+@plugin.include
+@highlights_group.child
+@crescent.command(name="list")
+async def list(ctx: Context) -> None:
+    if not await User.exists(user_id=ctx.user.id):
+        await ctx.respond("You do not have any highlights.")
+        return
+
+    user = await User.fetch(user_id=ctx.user.id)
```
CircuitSacul, line 68:
```py
if not (user := await User.exists(user_id=ctx.user.id)):
    await ctx.respond("You do not have any highlights.")
    return
```

CircuitSacul (changes requested):

mcodingbot/plugins/highlights.py
Lunarmagpie: `.append` is definitely the right one here

CircuitSacul: highlights_cache[word].extend(user_ids), instead of a for-loop

CircuitSacul:
> Investigate typing with fetchmany

I assume you're referring to the fact that the return types for ManyToMany are broken. If you update apgorm in this branch too, that will be fixed.

Lunarmagpie:
> I assume you're referring to the fact that the return types for ManyToMany are broken. If you update apgorm in this branch too, that will be fixed.

Yes, I checked the source code locally and there was a problem with apgom. Was up to date tho.

mcodingbot/plugins/highlights.py
Lunarmagpie: I think using the walrus operator when you aren't using the variable inside if the if statement is an antipattern tbh

CircuitSacul:
> > I assume you're referring to the fact that the return types for ManyToMany are broken. If you update apgorm in this branch too, that will be fixed.
> 
> Yes, I checked the source code locally and there was a problem with apgom. Was up to date tho. Just noticed you made a new release

I just now made a release. The fix has been sitting in the repo, but I hadn't pushed it to pypi yet.

mcodingbot/plugins/highlights.py
CircuitSacul: I don't think so, it's how a lot of people use it (sourcery also recommends this pattern)

mcodingbot/database/models/user.py
Lunarmagpie: the typehints in the static method are technically wrong without it

CircuitSacul: how so?

Lunarmagpie:
```python
class A:
    @staticmethod
     def f() -> A"
 
class B(A):
    ...
```
B now return A when you call the static method and thats not expected. So just say you can't subclass at all because you aren't supposed to subclass A.

mcodingbot/plugins/highlights.py
Lunarmagpie: the walrus greatly increases cognitive complexity

CircuitSacul: Then do ```py
user = await User.exists(user_id=ctx.user.id)
if not user:
    await ctx.respond(...)
    return
```

Comment (Lunarmagpie) [reply mcodingbot/plugins/highlights.py:68]: thats what im doing. I think its much more readable, and especially much more beginner friendly.

**Review by trag1c — CHANGES_REQUESTED** [review]

File: mcodingbot/plugins/highlights.py [hunk 10]
```diff
@@ -0,0 +1,114 @@
...
+        await ctx.respond("You do not have any highlights.")
+        return
+
+    words = await user.words.fetchmany()
+
+    if not words:
+        await ctx.respond("You do not have any highlights.")
+        return
+
+    await ctx.respond("\n".join((word.word for word in words)))  # type: ignore
```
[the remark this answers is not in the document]
Lunarmagpie: im intending to replace this with an embed fyi. i didnt write it though cause im not an artist.

mcodingbot/database/models/user.py
Lunarmagpie: I don't understand how I can do that here

mcodingbot/database/models/user.py
Lunarmagpie: Ill remove the final decorator

mcodingbot/plugins/highlights.py
Lunarmagpie: I added descriptions. Feel free to suggest changes.

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+@plugin.include
+@crescent.event
+async def on_start(_: hikari.StartingEvent) -> None:
+    words = await Word.fetchmany()
+
+    for word, users in zip(
+        words,
+        await asyncio.gather(*(word.users.fetchmany() for word in words)),
```
CircuitSacul, line 108: Since you only get the user_id, it would be more efficient to query the UserWord model instead.

```py
for word, user_words in zip(
    words,
    await asyncio.gather(*(UserWord.fetchmany(word_id=word.id) for word in words))),
):
    _cache_highlight(word.word, *(user_word.user_id for user_word in user_words))
```

CircuitSacul (changes requested):

Lunarmagpie:
Only visuals are left. I don't have any ideas and don't really feel like spending my time implementing anything so someone else can do it.

CircuitSacul:
> Only visuals are left. I don't have any ideas and don't really feel like spending my time implementing anything so someone else can do it.

I'll do that part then

CircuitSacul:
We should also make it ignore the person who owns the highlight. Like, if I have "1984" highlighted and then I say 1984, it shouldn't notify me.

Lunarmagpie:
Thats really important

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,161 @@
...
+    for highlight, users in highlights_cache.items():
+        if highlight in event.content.split():
+            for user_id in users:
+                asyncio.ensure_future(
+                    _dm_user_highlight(
+                        user_id=user_id,
+                        highlight=highlight,
+                        msg_link=event.message.make_link(event.guild_id),
+                    )
+                )
```
CircuitSacul, line 161:
```py
for user_id in users:
    if user_id == event.user_id:
        continue
```

might be event.author.id or event.user.id or event.message.author.id or something like that

Lunarmagpie: I think i completely misconceptualized this in my head. That should work.

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,163 @@
...
+async def on_message(event: hikari.GuildMessageCreateEvent) -> None:
+    if not event.content or event.is_bot:
+        return
+
+    for highlight, users in highlights_cache.items():
+        if highlight in event.content.split():
+            for user_id in users:
+                if user_id == event.author.id:
+                    continue
+                asyncio.ensure_future(
```
Lunarmagpie, line 157: Does anyone have strong feeling for `ensure_version` or `create_task` here?

mcodingbot/plugins/highlights.py
Lunarmagpie: Done

mcodingbot/plugins/highlights.py
CircuitSacul: uh... what's the difference 😅

Lunarmagpie: i don't really know

mcodingbot/plugins/highlights.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@plugin.include
+@crescent.event
+async def on_message(event: hikari.GuildMessageCreateEvent) -> None:
+    if not event.content or event.is_bot:
+        return
+
+    for highlight, users in highlights_cache.items():
+        if highlight in event.content.split():
```
CircuitSacul, line 169: Doing this means that if the word is followed by a comma, it will not trigger. I would just go with `highlight in event.content` for now. Regex support in the future anyways, I think?

CircuitSacul (changes requested):
Think this is no longer a draft too

mcodingbot/plugins/highlights.py
Lunarmagpie: I would prefer stripping commas before comparing

MithicSpirit: If regex won't be available now then some strings need to be updated to reflect that.

CircuitSacul: > I would prefer stripping commas before comparing

Any non-alphanumerica characters should be stripped. Here's a list I can think of: ```
`'";:,.<>?/\|[]{}()*&^%$@!
```

The problem is, this means that highlights that rely on non-alphanumeric characters won't work right. I really think we should just remove the split.

Comment (CircuitSacul) [reply mcodingbot/plugins/highlights.py:169]: > If regex won't be available now then some strings need to be updated to reflect that.

I didn't see any, maybe go through and add comments to those?

Comment (Lunarmagpie) [reply mcodingbot/plugins/highlights.py:169]: > I didn't see any, maybe go through and add comments to those?


I see them. I can go fix it.

Comment (Lunarmagpie) [reply mcodingbot/plugins/highlights.py:169]: > The problem is, this means that highlights that rely on non-alphanumeric characters won't work right. I really think we should just remove the split.

It might just be less work to add regex support now

Comment (MithicSpirit) [reply mcodingbot/plugins/highlights.py:169]: On second thought we should definitely remove the split, otherwise multi-word highlights don't work.

Comment (trag1c) [reply mcodingbot/plugins/highlights.py:169]: > > I would prefer stripping commas before comparing
> 
> Any non-alphanumerica characters should be stripped. Here's a list I can think of:
> 
> ```
> `'";:,.<>?/\|[]{}()*&^%$@!
> ```

So just `string.punctuation` maybe?

Comment (trag1c) [reply mcodingbot/plugins/highlights.py:169]:
```pycon
>>> import string
>>> print(string.punctuation)
!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~
```

CircuitSacul: @trag1c very helpful thanks.

Still, that leaves us with the problem of having highlights that include punctuation.

MithicSpirit: If we don't want to deal with implementing the regex timeout code now we could use escape the highlight and then wrap it in `\b`s, then match on that with regex. It's a simple enough pattern that it shouldn't have performance issues.

MithicSpirit: > Still, that leaves us with the problem of having highlights that include punctuation.

or (and more importantly) spaces.

CircuitSacul: > If we don't want to deal with implementing the regex timeout code now we could use escape the highlight and then wrap it in \bs, then match on that with regex. It's a simple enough pattern that it shouldn't have performance issues.

won't that just leave us back with the problem of `test` not matchint `test.`?

MithicSpirit: > won't that just leave us back with the problem of `test` not matchint `test.`?

I'm pretty sure that between a letter and a period counts as a word boundary, so the regex `\btest\b` (which corresponds to the highlight `test`) should match `test.`.

The only issue I can see with this is properly escaping the input string.

CircuitSacul: All of this will be solved once we have regex highlights I think. But for now, I think using `highlight in message.content` allows more control over what's matched and what isn't than doing `highlight in message.content.split()`. Neither are completely ideal, but again, we can fix that with regex highlights.

Lunarmagpie: Doing just `in` for now is the easiest I guess

CircuitSacul: Sounds good (test comment)

Lunarmagpie (commented):
lgtm

CircuitSacul (approved):

MithicSpirit (approved):
Don't know enough about ORMs and DBs to review that part but everything else LGTM.

trag1c (approved):

PR #75, opened 2022-09-30
Put stat names after the values

trag1c:
Changes `Cakes: 2**4.68 (25)` to `2**4.68 (25) cakes` to avoid stuff like this:

image

It doesn't look the best, but I don't have a better idea 💀


Lunarmagpie (approved):

qexat:
Hmm couldn't it make worse? For instance, if the numbers get pretty big, it would even hide the label meaning that we couldn't even know that this would be the members count, right?

CircuitSacul (approved):

Lunarmagpie:
> Hmm couldn't it make worse? For instance, if the numbers get pretty big, it would even hide the label meaning that we couldn't even know that this would be the members count, right?

We can revert if it looks bad

PR #86, opened 2022-11-17
Highlight Cooldowns

CircuitSacul:
This makes a few changes to improve highlights
 - [x] If a message contains multiple triggers, combine them into a single DM.
 - [x] Have a cooldown of 30s per (highlight, user, channel)
 - [x] If a user sends a message in a channel, don't send them any highlights for the next minute.

These are just my feel for how it should work. If you think it should be tweaked lmk.

### DIFF
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -25,7 +25,7 @@ class Config:
     donor_role: int = -1
 
     # capacity, cooldown
-    message_sent_cooldown: tuple[float, float] = (1, 60)
+    highlight_message_sent_cooldown: tuple[float, float] = (1, 60)
     highlight_trigger_cooldown: tuple[float, float] = (1, 30)
 
     mcoding_yt_id: str = "YOUTUBE_CHANNEL_ID"
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -24,6 +24,9 @@ class Config:
     patron_role: int = -1
     donor_role: int = -1
 
+    # capacity, cooldown
+    message_sent_cooldown: tuple[float, float] = (1, 60)
+
     mcoding_yt_id: str = "YOUTUBE_CHANNEL_ID"
     yt_api_key: str = "YOUTUBE_API_KEY"
 
--- a/mcodingbot/config.py
+++ b/mcodingbot/config.py
@@ -26,6 +26,7 @@ class Config:
 
     # capacity, cooldown
     message_sent_cooldown: tuple[float, float] = (1, 60)
+    highlight_trigger_cooldown: tuple[float, float] = (1, 30)
 
     mcoding_yt_id: str = "YOUTUBE_CHANNEL_ID"
     yt_api_key: str = "YOUTUBE_API_KEY"
--- a/mcodingbot/plugins/highlights.py
+++ b/mcodingbot/plugins/highlights.py
@@ -207,10 +207,7 @@ async def on_message(event: hikari.GuildMessageCreateEvent) -> None:
                 highlights[user_id].append(highlight)
 
     for user_id, hls in highlights.items():
-        if (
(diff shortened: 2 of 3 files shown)


MithicSpirit:
These changes sound good to me, though please make the two cooldowns relatively easy to configure (like `MAX_HIGHLIGHTS` and `MAX_HIGHLIGHT_LENGTH`).

Might be interesting for us to look into having a proper config file for these things in the future (maybe we can commandeer the `config.json` for these purposes?).

CircuitSacul:
> These changes sound good to me, though please make the two cooldowns relatively easy to configure (like `MAX_HIGHLIGHTS` and `MAX_HIGHLIGHT_LENGTH`).
> 
> Might be interesting for us to look into having a proper config file for these things in the future (maybe we can commandeer the `config.json` for these purposes?).

Well these things would undoubtedly be defined in config.json (with defaults in config.py). Unless you meant per-user configuration? I think that's possible too.

Lunarmagpie:
> Well these things would undoubtedly be defined in config.json (with defaults in config.py). Unless you meant per-user configuration? I think that's possible too.

I think moving all config options to `config.py` would be a good idea. May open a PR for that before this one gets merged.

MithicSpirit:
> Well these things would undoubtedly be defined in config.json (with defaults in config.py). Unless you meant per-user configuration? I think that's possible too.

`MAX_HIGHLIGHTS` and `MAX_HIGHLIGHT_LENGTH` aren't in `config.json` :p. As for per-user configuration, it would be nice but probably overkill for now.

CircuitSacul:
> `MAX_HIGHLIGHTS` and `MAX_HIGHLIGHT_LENGTH` aren't in `config.json` :p. As for per-user configuration, it would be nice but probably overkill for now.

Oh oof. They should've been. Well, I'll be putting these settings in config. And yeah, per-user config is outside the scope of this pr anyways

CircuitSacul:
This is done. Ready for review.

mcodingbot/plugins/highlights.py
```diff
@@ -166,18 +173,49 @@ async def _dm_user_highlight(
...
-                        user_id=user_id,
-                    )
-                )
+                highlights[user_id].append(highlight)
+
+    for user_id, hls in highlights.items():
+        if (
+            sent_message_cooldown.get_retry_after((user_id, event.channel_id))
+            != 0
+        ):
```
Lunarmagpie, line 213: I think the `!= 0` can be removed. It might fit on one line then.

mcodingbot/config.py
```diff
@@ -24,6 +24,10 @@ class Config:
     patron_role: int = -1
     donor_role: int = -1
 
+    # capacity, cooldown
+    message_sent_cooldown: tuple[float, float] = (1, 60)
```
Lunarmagpie, line 28: I think it is confusing what the `message_sent_cooldown` is for
```suggestion
    highlight_message_sent_cooldown: tuple[float, float] = (1, 60)
```

Lunarmagpie (changes requested):
I think switching to `NamedTuple` for buckets would help reduce mistakes when editing this code in the future.

Lunarmagpie:
Still would prefer if `NamedTuple` was used.

Lunarmagpie (approved):

Lunarmagpie (changes requested):
Actually the bucket should probably be `SentMessageBucket` so its obvious what variable its used with

CircuitSacul:
> Actually the bucket should probably be `SentMessageBucket`

I just named it "UserChannelBucket" b/c that's what it is literally

Lunarmagpie:
> I just named it "UserChannelBucket" b/c that's what it is literally

No reason to explain the properties of the class. Makes more sense to name it so its clear where its being used.

Lunarmagpie (approved):

trag1c (approved):

PR #87, opened 2022-11-18
Add `poetry update` to restart script

Lunarmagpie:
Dependencies should be updated when the bot restarts since there isn't a lockfile in the repo

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -3,6 +3,7 @@
 while true
 do
     git pull
+    "${1:-python}" -m poetry update
     "${1:-python}" -m poetry install
     "${1:-python}" -m poetry run python -m mcodingbot
 


run.sh
```diff
@@ -3,6 +3,7 @@
 while true
 do
     git pull
+    "${1:-python}" -m poetry update
     "${1:-python}" -m poetry install
```
CircuitSacul, line 7: I don't think we need install, since there's update.

MithicSpirit: We should probably leave it in unless someone tests it and confirms one way or the other.

Lunarmagpie:
I dont think `poetry update` installs the mcodingbot package, but idk if that matters.

MithicSpirit (commented):
Should we update before or after installing?

Lunarmagpie:
Before, otherwise `install` would install outdated dependencies.

MithicSpirit:
I am still of the opinion we should keep a lockfile on some stable branch and remove it on a dev branch (which is where pull requests are made to), but whatever.

CircuitSacul:
> I am still of the opinion we should keep a lockfile on some stable branch (which is where the bot pulls from) and remove it on a dev branch (which is where pull requests are made to), but whatever.

I'm starting to agree with you, tbh, but I think this is good for now.

Lunarmagpie:
> I am still of the opinion we should keep a lockfile on some stable branch and remove it on a dev branch (which is where pull requests are made to), but whatever.

No reason to use poetry otherwise. We could migrate to setuptools with the current setup and it would be easier for new contributors lol.

CircuitSacul (approved):

MithicSpirit (approved):
I don't approve of removing the lockfile but whatever

Lunarmagpie:
> I don't approve of removing the lockfile but whatever

The lockfile was gone anyway for some reason?

PR #88, opened 2022-11-18
Replace errors on bot launch with warnings

Lunarmagpie:
This PR replaces the exceptions when the bot launches with warnings.

### DIFF
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -34,6 +34,25 @@ class Bot(crescent.Bot):
                 "Server stats and donor roles will not be updated because"
                 " the mcoding server id is not provided. Is this intended?"
             )
+        else:
+            if not CONFIG.sub_count_channel:
+                _LOG.warning(
+                    "Sub count channel id for mcoding required to post stats."
+                )
+            if not CONFIG.view_count_channel:
+                _LOG.warning(
+                    "View count channel id for mcoding required to post stats."
+                )
+            if not CONFIG.member_count_channel:
+                _LOG.warning(
+                    "Member count channel id for mcoding required to post"
+                    " stats."
+                )
+            if not CONFIG.donor_role:
+                _LOG.warning(
+                    "Doner role id for mcoding server expected to update"
+                    "donor roles."
+                )
 
     @property
     def session(self) -> aiohttp.ClientSession:
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -36,7 +36,7 @@ class Bot(crescent.Bot):
         if not CONFIG.mcoding_server:
             _LOG.warning(
                 "Server stats and donor roles will not be updated because"
-                " the mcoding server id is not provided. Is this intended?"
+                " `mcoding_server` is not provided. Is this intended?"
             )
         else:
             if not CONFIG.sub_count_channel:
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -29,6 +29,12 @@ class Bot(crescent.Bot):
         self._session: aiohttp.ClientSession | None = None
         self._db: Database | None = None
 
+        if not CONFIG.mcoding_server:
+            _LOG.warning(
+                "Server stats and donor roles will not be updated because"
+                " the mcoding server id is not provided. Is this intended?"
+            )
+
     @property
     def session(self) -> aiohttp.ClientSession:
         if not self._session:
--- a/mcodingbot/bot.py
+++ b/mcodingbot/bot.py
@@ -37,21 +37,23 @@ class Bot(crescent.Bot):
         else:
             if not CONFIG.sub_count_channel:
                 _LOG.warning(
-                    "Sub count channel id for mcoding required to post stats."
+                    "Sub count channel id for mcoding server required to post"
+                    " stats."
                 )
             if not CONFIG.view_count_channel:
                 _LOG.warning(
-                    "View count channel id for mcoding required to post stats."
+                    "View count channel id for mcoding server required to post"
+                    " stats."
                 )
             if not CONFIG.member_count_channel:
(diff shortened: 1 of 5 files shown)


mcodingbot/plugins/donors.py
```diff
@@ -82,6 +88,15 @@ async def _is_donor(member: hikari.Member) -> bool:
 async def _update_donor_role(
     member: int | hikari.Member, is_donor: bool
 ) -> None:
+    if not CONFIG.mcoding_server:
+        LOGGER.warning(
+            "Could not update donor roles because mcoding server id was not"
+            " provieded. Is this intended?"
+        )
```
CircuitSacul, line 95: Rather than warning here, maybe it should be done in the start function of the bot? Here it should just return. Thinking this because right now, this only warns if you use the set-donor-status command, but not for the loop (because the loop/task already returns)

mcodingbot/plugins/stats.py
```diff
@@ -30,6 +30,23 @@ async def loop() -> None:
 
 
 async def update_channels(bot: Bot) -> None:
+    if not CONFIG.mcoding_server:
+        LOGGER.warning(
+            "Could not update channel stats because mcoding server id was not"
+            " provided. Is this intended?"
+        )
```
CircuitSacul, line 37: Same thing here

CircuitSacul (changes requested):

mcodingbot/plugins/donors.py
```diff
@@ -82,6 +88,11 @@ async def _is_donor(member: hikari.Member) -> bool:
 async def _update_donor_role(
     member: int | hikari.Member, is_donor: bool
 ) -> None:
+    if not CONFIG.mcoding_server:
+        return
+
+    assert CONFIG.donor_role
```
CircuitSacul, line 94: Just return here too.

mcodingbot/plugins/stats.py
```diff
@@ -30,8 +30,15 @@ async def loop() -> None:
...
 
 async def update_channels(bot: Bot) -> None:
+    if not CONFIG.mcoding_server:
+        return
+
     stats = await get_stats(bot)
 
+    assert CONFIG.sub_count_channel
+    assert CONFIG.view_count_channel
+    assert CONFIG.member_count_channel
```
CircuitSacul, line 40: Just only update the ones that exist.

CircuitSacul (changes requested):
Instead of asserting the config values, returning is fine. This way you can specify a guild but not the other things, and still only get logger warnings.

CircuitSacul (approved):

mcodingbot/bot.py
```diff
@@ -29,6 +29,33 @@ def __init__(self) -> None:
...
+            if not CONFIG.member_count_channel:
+                _LOG.warning(
+                    "Member count channel id for mcoding server required to"
+                    " post stats."
+                )
+            if not CONFIG.donor_role:
+                _LOG.warning(
+                    "Doner role id for mcoding server expected to update"
+                    " donor roles."
+                )
```
trag1c, line 57: Can't you DRY this? Also "Doner role" 🇹🇷

mcodingbot/config.py
```diff
@@ -17,12 +17,12 @@ class Config:
...
-    view_count_channel: int = -1
-    member_count_channel: int = -1
-    patron_role: int = -1
-    donor_role: int = -1
+    mcoding_server: int | None = None
+    sub_count_channel: int | None = None
+    view_count_channel: int | None = None
+    member_count_channel: int | None = None
+    patron_role: int | None = None
+    donor_role: int | None = None
```
trag1c, line 25: Why?

mcodingbot/plugins/donors.py
```diff
@@ -82,6 +88,12 @@ async def _is_donor(member: hikari.Member) -> bool:
 async def _update_donor_role(
     member: int | hikari.Member, is_donor: bool
 ) -> None:
+    if not CONFIG.mcoding_server:
+        return
+
+    if not CONFIG.donor_role:
+        return
+
```
trag1c, line 96:
```suggestion
    if not (CONFIG.mcoding_server and CONFIG.donor_role):
        return
```

trag1c (changes requested):

mcodingbot/config.py
Lunarmagpie: The config options are optional so they default to `None`. Defaulting to `-1` like they did before makes it hard to check if a value was supplied.

mcodingbot/bot.py
Lunarmagpie: Idk a way to make it DRYer and keep it as readable

CircuitSacul: Now that I read these I feel like it should match the variable name exactly ("view_count_channel_id" not "View count channel id")

Lunarmagpie: I tried drying this and I made the changes circuit requested

trag1c (approved):

CircuitSacul (approved):

trag1c:
@CircuitSacul can you maybe invert the logic so that there's not this much indented code

CircuitSacul:
> @CircuitSacul can you maybe invert the logic so that there's not this much indented code

I don't see how, since each section is independent of the other

trag1c:
> > @CircuitSacul can you maybe invert the logic so that there's not this much indented code
> 
> I don't see how, since each section is independent of the other

lines 54–77 in stats.py

trag1c:
fix typos in the new comments

CircuitSacul:
> fix typos in the new comments

oops. fixed

trag1c (approved):

Lunarmagpie:
I dont think we 100% needs the comments. LGTM though.

PR #89, opened 2022-11-24
Made stats in thousands truncate to 1 decimal place

trag1c:
Numbers like `1357` were displayed as `1.35K`, which is a bruh moment because it's literally longer than just displaying the full number itself. So from now on it's gonna get truncated to just `1.3K`.

### DIFF
--- a/mcodingbot/plugins/stats.py
+++ b/mcodingbot/plugins/stats.py
@@ -143,7 +143,9 @@ def display_stats(stat: int | float) -> str:
         pretty_stat = stat / 1_000_000
         unit = "M"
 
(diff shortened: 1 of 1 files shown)


Lunarmagpie (approved):
Seems reasonable to me

CircuitSacul:
Why wouldn't you want to truncate if the unit was "M"?

CircuitSacul:
Probably better to truncate to 1 if the unit is non-empty

trag1c:
> Why wouldn't you want to truncate if the unit was "M"?

1. `1.23M` is still shorter than `1234567`; for `K` that's not really the case
2. I'd rather see a stat bump every 10k views than 100k views

CircuitSacul (approved):

PR #90, opened 2022-11-27
Exact stats values command

FuexFollets:
A command `/stats` that returns exact statistics for subscribers, views, and server members

### DIFF
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -33,7 +33,7 @@ async def links(ctx: Context) -> None:
 @plugin.include
 @crescent.command(name="stats", description="Exact values for mCoding statistics")
 async def stats(ctx: Context) -> None:
-    stats = await get_stats(ctx.session.bot)
+    stats = await stats.get_stats(ctx.session.bot)
     embed = hikari.Embed(
         title="mCoding stats",
         color=CONFIG.theme,
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -1,8 +1,11 @@
+from __future__ import annotations
+
 import crescent
 import hikari
 
 from mcodingbot.config import CONFIG
 from mcodingbot.utils import Context, Plugin
+from mcodingbot.plugins import stats
 
 plugin = Plugin()
 
@@ -26,3 +29,19 @@ async def links(ctx: Context) -> None:
         ),
     )
     await ctx.respond(embed=embed)
+
+@plugin.include
+@crescent.command(name="stats", description="Exact values for mCoding statistics")
+async def stats(ctx: Context) -> None:
+    stats = await get_stats(ctx.session.bot)
+    embed = hikari.Embed(
+        title="mCoding stats",
+        color=CONFIG.theme,
+        description=(
+            f"Server members: `{ctx.server.member_count:,}`\n"
+            f"Subscribers: `{stats.subs:,}`\n"
+            f"Views: {stats.views:,}"
+        ),
+    )
+    await ctx.respond(embed=embed)
+
--- a/mcodingbot/plugins/info.py
+++ b/mcodingbot/plugins/info.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import crescent
 import hikari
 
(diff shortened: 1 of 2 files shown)


mcodingbot/plugins/stats.py
```diff
@@ -150,3 +153,21 @@ def display_stats(stat: int | float) -> str:
...
 
     return f"2**{exp_stat} ({pretty_stat}{unit})"
+
+
+@plugin.include
+@crescent.command(
+    name="stats", description="Exact values for mCoding statistics"
+)
+async def stats(ctx: Context) -> None:
+    stats = await get_stats(ctx.bot)
```
Lunarmagpie, line 163: Stats should not be fetched every time the command is used. They should be cached when the function called `loop` is run and fetched later. Is is too prevent API ratelimits.

Lunarmagpie (changes requested):

mcodingbot/plugins/info.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
```
CircuitSacul, line 2: Although this should probably be done at some point, it's outside the scope of this PR.

mcodingbot/plugins/stats.py
CircuitSacul: I believe you can use the _last_known_stats variable (it's a global in this file)

CircuitSacul: Also I think the command should be at the top of the file rather than the bottom. cc @Lunarmagpie @trag1c

CircuitSacul (changes requested):

mcodingbot/plugins/stats.py
Lunarmagpie: Agreed

trag1c: I don't mind to be fair

FuexFollets:
Ive applied the reccomendatioms from the comments in the pr

trag1c:
You removed the wrong future import. Circuit was talking about the one you added in `info.py`, and you removed the one found in `stats.py`

mcodingbot/plugins/stats.py
```diff
@@ -20,6 +20,23 @@
...
+    name="stats", description="Exact values for mCoding statistics"
+)
+async def stats(ctx: Context) -> None:
+    embed = hikari.Embed(
+        title="mCoding stats",
+        color=CONFIG.theme,
+        description=(
+            f"Server members: `{ctx.server.member_count:,}`\n"
+            f"Subscribers: `{_last_known_stats.subs:,}`\n"
+            f"Views: {_last_known_stats.views:,}"
```
trag1c, line 34: Also, you didn't put backticks for the last one
```suggestion
            f"Server members: `{ctx.server.member_count:,}`\n"
            f"Subscribers: `{_last_known_stats.subs:,}`\n"
            f"Views: `{_last_known_stats.views:,}`"
```

trag1c (changes requested):

mcodingbot/plugins/stats.py
```diff
@@ -20,6 +20,23 @@
...
+@plugin.include
+@crescent.command(
+    name="stats", description="Exact values for mCoding statistics"
+)
+async def stats(ctx: Context) -> None:
+    embed = hikari.Embed(
+        title="mCoding stats",
+        color=CONFIG.theme,
+        description=(
+            f"Server members: `{ctx.server.member_count:,}`\n"
```
Lunarmagpie, line 32: None of this exists. Please run your code before making a PR.

CircuitSacul: @FuexFollets you can use `_last_known_stats.subs` and `_last_known_stats.views` for sub and view counts.

For the member count, lines 71-79 of this file are how it should be done. I would make that a function in this file to avoid code duplication.

FuexFollets:
Ive added the member count to the `_last_known_stats` global variable. Any thoughts? It just made sense

CircuitSacul:
> Ive added the member count to the `_last_known_stats` global variable. Any thoughts? It just made sense

Actually yeah I think that's perfect.

CircuitSacul (approved):

trag1c (approved):

Lunarmagpie (approved):

PR #92, opened 2022-11-28
Changed Stats to use ints

### LINKED ISSUE #91: Subscriber stats and views should be int rather than float
Floats are a bit unreasonable for these discrete values

trag1c:
Closes #91


Lunarmagpie (approved):
`display_stats` needs to convert to a float because we don't want integer division.
Edit: did not mean to approve

trag1c:
> `display_stats` needs to convert to a float because we don't want integer division.

w h a t
keep in mind this is Python

Lunarmagpie (approved):

CircuitSacul (approved):

PR #93, opened 2022-11-29
convert str result from yt api to int

CircuitSacul:
Fixes stats module.


trag1c (approved):

CircuitSacul:
I think I'm gonna commit mod abuse and merge this without another review

PR #94, opened 2022-12-02
Support ruff, remove flake8, and fix nox versioning, config in pyproject.toml

Lunarmagpie:
This PR cleans up CI by doing the stuff in the title


MithicSpirit:
We should add more python versions to the CI imo.

mypy.ini
```diff
@@ -1,11 +0,0 @@
-[mypy]
-warn_return_any=True
-warn_unused_configs=True
-strict=True
-namespace_packages=True
-
-[mypy-nox.*]
-ignore_missing_imports=True
```
MithicSpirit, line 8: What happened to `ignore-missing-imports` for the nox mypy module?

.isort.cfg
```diff
@@ -1,3 +0,0 @@
-[settings]
-profile = black
-line_length = 79
```
MithicSpirit, line 3: What happened to the line length limit in the isort configuration?

CircuitSacul:
> We should add more python versions to the CI imo, just to make sure that every version works correctly.

linting is going to be the same regardless of python version. For type-checking, aren't 3.* python versions backwards compatible?

.isort.cfg
Lunarmagpie: Pretty sure isn't needed because `profile=black` sets it to the right one

mypy.ini
Lunarmagpie: It wasn't doing anything. Nox is typed so im not sure why that was there to begin with.

.isort.cfg
CircuitSacul: > Pretty sure isn't needed because `profile=black` sets it to the right one

Black uses 88 by default

mypy.ini
CircuitSacul: CI passes so it clearly has types

.isort.cfg
Lunarmagpie: It copies the line length for black. I just checked and isort complains about a length 83 line with the config i set up.

CircuitSacul: Oh nice, so it actually looks for the black config.

Lunarmagpie:
> linting is going to be the same regardless of python version. For type-checking, aren't 3.* python versions backwards compatible?

Agree with this. I think we a script to import `mCodingBot` or tests to check that it actually works on multiple python version.

MithicSpirit (approved):
lgtm.

trag1c (approved):

PR #100, opened 2022-12-06
Use dependency groups and ignore ruff versions

Lunarmagpie:
This PR properly locks dependency versions in our CI


Lunarmagpie:
I can't merge this. Poetry is too slow :sob:

PR #103, opened 2022-12-08
Add cooldown between sending identical pep messages

Lunarmagpie:
This PR does what it says in the title. All details are documented in docstrings.

### DIFF
--- a/mcodingbot/plugins/peps.py
+++ b/mcodingbot/plugins/peps.py
@@ -15,7 +15,7 @@ from mcodingbot.config import CONFIG
 from mcodingbot.utils import Context, PEPInfo, PEPManager, Plugin
 
 
-class PepBucket(NamedTuple):
+class PEPBucket(NamedTuple):
     pep: int
     channel: int
 
@@ -38,17 +38,17 @@ recent_pep_responses: TTLCache[int, MessageInfo] = TTLCache(
 )
 plugin = Plugin()
 pep_manager = PEPManager()
-pep_cooldown: FixedMapping[PepBucket] = FixedMapping(*CONFIG.pep_cooldown)
+pep_cooldown: FixedMapping[PEPBucket] = FixedMapping(*CONFIG.pep_cooldown)
 
 
 def trigger_cooldowns(peps: Iterable[PEPInfo], channel_id: int) -> None:
     for pep in peps:
-        pep_cooldown.trigger(PepBucket(pep=pep.number, channel=channel_id))
+        pep_cooldown.trigger(PEPBucket(pep=pep.number, channel=channel_id))
 
 
 def reset_cooldowns(peps: Iterable[PEPInfo], channel_id: int) -> None:
     for pep in peps:
-        pep_cooldown.reset(PepBucket(pep=pep.number, channel=channel_id))
+        pep_cooldown.reset(PEPBucket(pep=pep.number, channel=channel_id))
 
 
 def filter_can_send(peps: list[PEPInfo], channel_id: int) -> Iterable[PEPInfo]:
@@ -59,7 +59,7 @@ def filter_can_send(peps: list[PEPInfo], channel_id: int) -> Iterable[PEPInfo]:
 
     def is_sendable(pep: PEPInfo) -> bool:
         return pep_cooldown.can_trigger(
-            PepBucket(pep=pep.number, channel=channel_id)
+            PEPBucket(pep=pep.number, channel=channel_id)
         )
 
     return filter(is_sendable, peps)
@@ -147,7 +147,7 @@ def get_peps_embed(refs: list[PEPInfo]) -> hikari.UndefinedOr[hikari.Embed]:
     if not refs:
         return hikari.UNDEFINED
 
-    pep_links_message = "\n".join(map(str, pep))
+    pep_links_message = "\n".join(map(str, refs))
 
     embed = hikari.Embed(description=pep_links_message, color=CONFIG.theme)
 
--- a/mcodingbot/plugins/peps.py
+++ b/mcodingbot/plugins/peps.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import re
 from contextlib import suppress
 from datetime import datetime, timedelta, timezone
-from typing import Iterable, NamedTuple
+from typing import Iterable, NamedTuple, Collection
 
 import crescent
 import hikari
@@ -22,7 +22,7 @@ class PEPBucket(NamedTuple):
 
 class MessageInfo(NamedTuple):
     message: int
-    peps: list[PEPInfo]
+    peps: set[PEPInfo]
 
 
 PEP_REGEX = re.compile(
@@ -34,7 +34,7 @@ MAX_AGE_FOR_SEND = timedelta(minutes=1)
 MAX_AGE_FOR_EDIT = timedelta(minutes=5)
 
 recent_pep_responses: TTLCache[int, MessageInfo] = TTLCache(
-    maxsize=100, ttl=MAX_AGE_FOR_EDIT.total_seconds()
+    maxsize=10_000, ttl=MAX_AGE_FOR_EDIT.total_seconds()
 )
 plugin = Plugin()
 pep_manager = PEPManager()
@@ -51,7 +51,9 @@ def reset_cooldowns(peps: Iterable[PEPInfo], channel_id: int) -> None:
         pep_cooldown.reset(PEPBucket(pep=pep.number, channel=channel_id))
 
 
-def filter_can_send(peps: list[PEPInfo], channel_id: int) -> Iterable[PEPInfo]:
+def filter_can_send(
+    peps: Iterable[PEPInfo], channel_id: int
+) -> Iterable[PEPInfo]:
     """
     Removes peps that can not be sent because of the cooldown from the list
     of peps.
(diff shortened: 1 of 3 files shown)


mcodingbot/plugins/peps.py
```diff
@@ -89,19 +127,36 @@ def within_age_cutoff(message_created_at: datetime) -> bool:
...
+
+def get_peps_embed(refs: list[PEPInfo]) -> hikari.UndefinedOr[hikari.Embed]:
+    """
+    Get a pep embed from a list of refs. If there are no refs,
+    return `hikari.UNDEFINED`.
+    """
+    if not refs:
+        return hikari.UNDEFINED
+
+    pep_links_message = "\n".join(str(pep) for pep in refs if pep)
```
Endercheif, line 155: Is the `if pep` at the end of the line needed because the function will always get a list with `PEPInfo`s?
```suggestion
    pep_links_message = "\n".join(str(pep) for pep in refs)
```

Endercheif (commented):
:)

mcodingbot/plugins/peps.py
```diff
@@ -3,14 +3,27 @@
...
+from floodgate import FixedMapping
 
 from mcodingbot.config import CONFIG
-from mcodingbot.utils import Context, PEPManager, Plugin
+from mcodingbot.utils import Context, PEPInfo, PEPManager, Plugin
+
+
+class PepBucket(NamedTuple):
+    pep: int
+    channel: int
```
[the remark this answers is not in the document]
Lunarmagpie: I chose `channel` because its consistent with bucket definitions in other files

mcodingbot/plugins/peps.py
```diff
@@ -3,14 +3,27 @@
...
 from cachetools import TTLCache
 from crescent.ext import tasks
+from floodgate import FixedMapping
 
 from mcodingbot.config import CONFIG
-from mcodingbot.utils import Context, PEPManager, Plugin
+from mcodingbot.utils import Context, PEPInfo, PEPManager, Plugin
+
+
+class PepBucket(NamedTuple):
```
trag1c, line 18: This name is not PEP8 compliant

mcodingbot/plugins/peps.py
```diff
@@ -89,19 +127,33 @@ def within_age_cutoff(message_created_at: datetime) -> bool:
...
+def get_pep_refs(content: hikari.UndefinedNoneOr[str]) -> list[PEPInfo]:
+    """
+    Return a sorted list of all the peps mentioned in a string.
+    """
+    if not content:
+        return []
+
+    peps = sorted(
+        [int(ref.group("pep")) for ref in re.finditer(PEP_REGEX, content)]
+    )
```
trag1c, line 139: Useless list comprehension + I don't see the point of `re.finditer` instead of using the pattern directly
```suggestion
    peps = sorted(int(ref.group("pep")) for ref in PEP_REGEX.finditer(content))
```

mcodingbot/plugins/peps.py
```diff
@@ -89,19 +127,33 @@ def within_age_cutoff(message_created_at: datetime) -> bool:
...
-        return None
+def get_peps_embed(refs: list[PEPInfo]) -> hikari.UndefinedOr[hikari.Embed]:
+    """
+    Get a pep embed from a list of refs. If there are no refs,
+    return `hikari.UNDEFINED`.
+    """
+    if not refs:
+        return hikari.UNDEFINED
+
+    pep_links_message = "\n".join(str(pep) for pep in refs)
```
trag1c, line 152:
```suggestion
    pep_links_message = "\n".join(map(str, pep))
```

mcodingbot/plugins/peps.py
```diff
@@ -89,19 +127,33 @@ def within_age_cutoff(message_created_at: datetime) -> bool:
...
+    Return a sorted list of all the peps mentioned in a string.
+    """
+    if not content:
+        return []
+
+    peps = sorted(
+        [int(ref.group("pep")) for ref in re.finditer(PEP_REGEX, content)]
+    )
+
+    return [pep for pep in map(pep_manager.get, peps) if pep]
```
trag1c, line 141: Maybe?
```suggestion
    return list(filter(None, map(pep_manager.get, peps)))
```

trag1c (changes requested):

mcodingbot/plugins/peps.py
Lunarmagpie: I just moved this didn't write it lol. Commiting your suggestion.

mcodingbot/plugins/peps.py
Endercheif: imo the original is better

mcodingbot/plugins/peps.py
```diff
@@ -89,19 +127,31 @@ def within_age_cutoff(message_created_at: datetime) -> bool:
...
+
+def get_peps_embed(refs: list[PEPInfo]) -> hikari.UndefinedOr[hikari.Embed]:
+    """
+    Get a pep embed from a list of refs. If there are no refs,
+    return `hikari.UNDEFINED`.
+    """
+    if not refs:
+        return hikari.UNDEFINED
+
+    pep_links_message = "\n".join(map(str, pep))
```
Endercheif, line 150:
```suggestion
    pep_links_message = "\n".join(map(str, refs))
```

mcodingbot/plugins/peps.py
Lunarmagpie: I prefer tragic's suggestion for style because it doesn't mix iterator functions and list comps. List comp more readable tho cause tbh I didn't even know this was a python feature until now so I'm not sure.

Endercheif (changes requested):

mcodingbot/plugins/peps.py
Lunarmagpie: real

mcodingbot/plugins/peps.py
Endercheif: will fix the failing check

trag1c: whoops im a reer

mcodingbot/plugins/peps.py
Endercheif: The only reason I prefer the original is due to readability.

mcodingbot/plugins/peps.py
Endercheif: he just like me frfr

mcodingbot/plugins/peps.py
Endercheif: I change my mind. I like trag1c

Endercheif (approved):

Endercheif:
@trag1c

Endercheif:
<@252206453878685697> approve changes

trag1c (approved):

Endercheif:
merging. any objections then type on discord?

mcodingbot/plugins/peps.py
```diff
@@ -20,11 +33,38 @@
...
+    Removes peps that can not be sent because of the cooldown from the list
+    of peps.
+    """
+
+    def is_sendable(pep: PEPInfo) -> bool:
+        return pep_cooldown.can_trigger(
+            PEPBucket(pep=pep.number, channel=channel_id)
+        )
+
+    return filter(is_sendable, peps)
```
Endercheif, line 67: Every instance of this function being called wraps it in a `set`.
Why not just return a set once instead.

```suggestion
def filter_can_send(
    peps: Iterable[PEPInfo], channel_id: int
) -> set[PEPInfo]:
    """
    Removes peps that can not be sent because of the cooldown from the list
    of peps.
    """

    def is_sendable(pep: PEPInfo) -> bool:
        return pep_cooldown.can_trigger(
            PEPBucket(pep=pep.number, channel=channel_id)
        )

    return set(filter(is_sendable, peps))
```

Endercheif (changes requested):

mcodingbot/plugins/peps.py
Lunarmagpie: The method is a filtering method. Not a convert to set method. Imo converting to a set may be too many things.

trag1c: that's not even a method 😶

CircuitSacul (approved):

mcodingbot/plugins/peps.py
CircuitSacul: personally I think it makes sense for it to return an Iterable. Even though it is always converted to a set, that's not because the return value should necessarily be a set. Kinda like if you have a function that calculates something, and the only thing you do with it is print the result - you might still want to just return the value, to keep functionality seperate.

Lunarmagpie: exactly my thoughts

trag1c (approved):

mcodingbot/utils/peps.py
```diff
@@ -122,3 +122,6 @@ def truncated_title(self) -> str:
 
     def __str__(self) -> str:
         return f"PEP {self.number}: [{self.title}]({self.link})"
+
+    def __hash__(self) -> int:
+        return self.number
```
Lunarmagpie, line 127: thinking a docstring could help a bit with clarity. May add that.

trag1c: no need imo, the method is simple enough

Endercheif (approved):
ball

PR #106, opened 2022-12-24
check permissions before sending highlights

circuitsacul:


mcodingbot/plugins/highlights.py
```diff
@@ -2,21 +2,45 @@
...
 
+if TYPE_CHECKING:
+    from mcodingbot.bot import Bot
+
 MAX_HIGHLIGHTS = 25
 MAX_HIGHLIGHT_LENGTH = 32
 
 
+async def has_permission(
+    bot: Bot, guild_id: int, user_id: int, channel_id: int
```
Lunarmagpie, line 25: No reason to pass bot in here. Can just do `plugin.bot` or `plugin.app` don't remember

mcodingbot/plugins/highlights.py
```diff
@@ -2,21 +2,45 @@
...
 import asyncio
 from collections import defaultdict
-from typing import NamedTuple
+from typing import TYPE_CHECKING, NamedTuple, cast
 
 import crescent
 import hikari
 from asyncpg import UniqueViolationError
 from floodgate import FixedMapping
+from toolbox.members import calculate_permissions
```
Lunarmagpie, line 11: should be `import toolbox` or `from toolbox import calculate_permissions`. I prefer the first one OFC.

mcodingbot/plugins/highlights.py
```diff
@@ -2,21 +2,45 @@
...
+    if not channel:
+        # must be a thread
+        thread = bot.rest.fetch_channel(channel_id)
+        assert isinstance(
+            thread, (hikari.GuildThreadChannel, hikari.GuildNewsThread)
+        )
+        return await has_permission(bot, guild_id, user_id, thread.parent_id)
+
+    permissions = calculate_permissions(member, channel)
+    return hikari.Permissions.VIEW_CHANNEL in permissions
```
Lunarmagpie, line 41: Variable name doesn't add clarity imo
```suggestion
    return hikari.Permissions.VIEW_CHANNEL in calculate_permissions(member, channel)
```

zunda-arrow (changes requested):

mcodingbot/plugins/highlights.py
CircuitSacul: Alternative is ```py
return hikari.Permissions.VIEW_CHANNEL in calculate_permissions(
    member, channel
)
```

Comment (Lunarmagpie) [reply mcodingbot/plugins/highlights.py:41]: still better. Also this is the final straw im upping the line length.

Comment (Lunarmagpie) [reply mcodingbot/plugins/highlights.py]: resolving cause of trag1c's vote

File: mcodingbot/plugins/highlights.py [hunk 4]
```diff
@@ -17,6 +18,24 @@
...
+    member = plugin.app.cache.get_member(guild_id, user_id)
+    if not member:
+        return False
+
+    channel = plugin.app.cache.get_guild_channel(channel_id)
+    if not channel:
+        # must be a thread
+        thread = plugin.app.rest.fetch_channel(channel_id)
+        assert isinstance(
+            thread, (hikari.GuildThreadChannel, hikari.GuildNewsThread)
```
Lunarmagpie, line 31: can you use `hikari.GuildThreadChannel` here? This `assert` scares me.

CircuitSacul: The only channel type that hikari doesn't cache entirely are threads - if this assert fails then it needs to error, because something is completely broken with the cache.

Lunarmagpie: I don't like that the assert relies on the implementation details of the cache to error/not error. Its not an error if some other type is retrieved from the cache anyway thats not one of those two.

mcodingbot/plugins/highlights.py
```diff
@@ -17,6 +18,22 @@
...
 
 
+async def has_permission(guild_id: int, user_id: int, channel_id: int) -> bool:
+    member = plugin.app.cache.get_member(guild_id, user_id)
+    if not member:
+        return False
+
+    channel = plugin.app.cache.get_guild_channel(channel_id)
+    if not channel:
+        # must be a thread
```
Lunarmagpie, line 28: i think this comment should either be removed or explain why. It doesn't add anything to the code that u can't already see rn.

mcodingbot/plugins/highlights.py
```diff
@@ -166,6 +183,12 @@ async def on_start(_: hikari.StartingEvent) -> None:
 async def _dm_user_highlight(
     triggering_message: hikari.Message, triggers: list[str], user_id: int
 ) -> None:
+    assert triggering_message.guild_id is not None
```
Lunarmagpie, line 186: Should we use a comment to explain why the assert is ok? Im not sure if thats something worth doing tbh.

zunda-arrow (changes requested):

mcodingbot/plugins/highlights.py
CircuitSacul: I don't think it's worth it

mcodingbot/plugins/highlights.py
```diff
@@ -17,6 +18,22 @@
...
 
+async def has_permission(guild_id: int, user_id: int, channel_id: int) -> bool:
+    member = plugin.app.cache.get_member(guild_id, user_id)
+    if not member:
+        return False
+
+    channel = plugin.app.cache.get_guild_channel(channel_id)
+    if not channel:
+        # must be a thread
+        thread = plugin.app.rest.fetch_channel(channel_id)
```
Lunarmagpie, line 29: I think this needs to gracefully handle `hikari.NotFoundError` for me to be confident in the code

zunda-arrow (changes requested):

zunda-arrow (approved):

mcodingbot/plugins/highlights.py
```diff
@@ -1,22 +1,54 @@
...
 MAX_HIGHLIGHT_LENGTH = 32
 
 
+async def has_permission(guild_id: int, user_id: int, channel_id: int) -> bool:
+    member = plugin.app.cache.get_member(guild_id, user_id)
+    if not member:
+        return False
+
+    channel = plugin.app.cache.get_guild_channel(channel_id)
+    if not channel:
```
trag1c, line 30: I'd consider using `:=` here

Lunarmagpie: i hate that here so i think thats why circuit didn't do that

CircuitSacul: eh I think it's fine either way

MithicSpirit:
Any reason why this is still a patch? Also why isn't there a proper description for those merge commits?

PR #108, opened 2023-01-03
Add `no_db_mode` to config.json.example, fix unhandled when `no_db_mode` is true

Lunarmagpie:
I guess this is a little out of scope of each other but i cba to make 2 separate PRs.

This is needed because `no_db_mode` being false by default adds an exception that kills the bot on launcher if a postgres database isn't running.

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -1,4 +1,5 @@
 {
-    "db_password": "DATABASE-PASSWORD",
-    "discord_token": "DISCORD-TOKEN"
+    "discord_token": "DISCORD-TOKEN",
+    "no_db_mode": true,
+    "db_password": "OPTIONAL-DATABASE-PASSWORD"
 }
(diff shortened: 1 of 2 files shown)


trag1c (approved):

CircuitSacul (approved):

MithicSpirit:
> All the other db uses were handled in old PRs. This was missed when the highlights feature was introduced.

so things like the following fail gracefully if the database isn't available? https://github.com/mcb-dev/mCodingBot/blob/8bb548c7754b3e0a42aa756168da3983c92dfda3/mcodingbot/plugins/highlights.py#L67

Lunarmagpie:
You will get an error message that the db isn't set up I believe.

MithicSpirit:
Would it be possible to tell hikari to just not enable anything in the module if `no_db_mode` is set? maybe by just not importing it?

Lunarmagpie:
definitely but it would require modules to be loaded individually

MithicSpirit:
I would prefer that but it depends on what everyone else thinks (CC: @CircuitSacul @trag1c). Otherwise lgtm.

MithicSpirit (approved):

trag1c:
idc

CircuitSacul:
I think changing how modules are loaded is outside the scope of this pr anyways.

PR #111, opened 2023-02-03
fix highlight loading

CircuitSacul:


trag1c (approved):

Lunarmagpie (changes requested):
Isn't the `db` property broken now? I think there's no reason to check if the `db` is none now that It will never be `None`. Either remove that or replace it with checking if the `db` is open.

Lunarmagpie (approved):

PR #115, opened 2023-02-28
Make highlights case insensitive

trag1c:
initially went with making them casefold during creation but reverted that since there already are highlights in different cases

also I think it's not that heavy to run `.casefold()` on the content every time (can change if needed tho)

### DIFF
--- a/mcodingbot/plugins/highlights.py
+++ b/mcodingbot/plugins/highlights.py
@@ -70,7 +70,6 @@ class CreateHighlight:
             )
             return
 
-        self.word = self.word.casefold()
         highlight_model = await Highlight.get_or_create(highlight=self.word)
         user = await User.get_or_create(user_id=ctx.user.id)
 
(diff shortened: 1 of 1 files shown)


zunda-arrow (approved):
Lgtm

circuitsacul (approved):

MithicSpirit (changes requested):
Should we not casefold the highlight *before* storing it in the database? That should make it a bit more performant by only having to casefold once per highlight. Would require some work for migration tho.

EDIT: or ig we can just not migrate and require users to manually update their highlights if they're not casefolded :trollface:. Only issue would be that we would need to keep some way to delete highlights that are not casefolded.

zunda-arrow:
> Should we not casefold the highlight _before_ storing it in the database? That should make it a bit more performant by only having to casefold once per highlight. Would require some work for migration tho.
> 
> EDIT: or ig we can just not migrate and require users to manually update their highlights if they're not casefolded :trollface:. Only issue would be that we would need to keep some way to delete highlights that are not casefolded.

too much effort ngl

